PR #172, opened 2026-06-09
feat(subsonic): media-session cover art via content:// FileProvider (v0.1.4, pending car test)

TheZupZup:
## What & why

PR #170 made Subsonic/Navidrome cover art show **inside the app** by storing a credential-free `subsonic-cover:<id>` reference and resolving the authenticated `getCoverArt` URL only at render time. But the **platform media session** (lock screen / Android Auto now-playing card) loads `MediaItem.artUri` itself — in a place that render-time resolver can't reach, and where a credentialed URL must never go — so Subsonic art was still absent there.

This adds now-playing media-session artwork for Subsonic/Navidrome **without weakening the privacy model**: Linthra fetches the cover itself and hands the session only a safe local `file:`.

## How

- **New `MediaArtworkCache`** (`lib/core/services/media_artwork_cache.dart`): turns a credential-free reference into a private cached image. It mints the authenticated `getCoverArt` URL on demand (live session salt+token), fetches the bytes itself, and writes them to an app-private file under the OS cache dir. The cache key — and therefore the filename — is a **SHA-256 of the credential-free reference**, never the username, salt, token, server URL, or auth query. The authenticated URL is used once to fetch and **never returned, persisted, or logged**; only `image/*` 2xx bodies are kept. Any failure yields `null` (never throws).
- **`LinthraAudioHandler`** gains a `MediaArtworkResolver` seam. `MediaItem.artUri` is now computed through `_sessionArtUri`: a platform-loadable cover (Jellyfin token-free http, a local `file:`) passes through unchanged; a credential-free reference becomes its cached local `file:` **or null** — so an unloadable reference, and crucially **never** a credentialed URL, reaches the session. The now-playing track kicks off at most one fetch per cover, **off the playback path**, and re-broadcasts on success; a failure leaves the card art-less and never blocks playback.
- **`main`** wires the cache to the live Subsonic session.

## Privacy / security invariants held

- ✅ Never put username/password/salt/token/API key or the authenticated `getCoverArt` URL into `MediaItem.artUri`.
- ✅ Never persist authenticated cover URLs in the catalog DB (`Track.artworkUri` stays `subsonic-cover:<id>`).
- ✅ Never log authenticated cover URLs (fetch swallows errors silently; the Android Auto trace is unchanged and category-only).
- ✅ No credential in cache filenames, media metadata, diagnostics, or DB fields.
- ✅ Cast unchanged — Subsonic cast still omits artwork; Jellyfin cast still sends its token-free URL.
- ✅ Safe artwork can't be generated → fall back to **no artwork**, never a leak.

## Scope

- No change to audio playback behavior.
- Jellyfin artwork and local embedded artwork pass through **unchanged**.
- The in-app PR #170 resolver (`artworkImageProvider` / `installArtworkReferenceResolver`) is **untouched**.
- No new storage permissions (app-private OS cache dir; `file:` loaded in-process by `audio_service`).
- No version bump, tags, or fdroiddata edits.

## Tests

New `media_artwork_cache_test.dart` + an extended `linthra_audio_handler_test.dart` cover:

- a `subsonic-cover:<id>` reference becomes a safe local `file:` holding the fetched bytes;
- the cache filename / returned URI carry no credential, server URL, or auth query (asserted explicitly);
- signed-out / failed / empty / throwing fetches fall back to `null` without writing a file or blocking playback;
- disk reuse and concurrent-fetch dedup;
- the now-playing item shows the local cover while the reference and credential never reach `artUri`;
- without a resolver, a `subsonic-cover:` reference is dropped (never leaked) from `artUri`;
- Jellyfin http art and local file art still pass through unchanged, and the resolver isn't even consulted for them;
- browse-tree containers drop an unresolved reference.

Full suite: **1674 passing**, `flutter analyze` clean, `dart format` clean, secret-scan clean.

Targeted to be safe for the `v0.1.4` hotfix.

https://claude.ai/code/session_01BLf35327oj8ArATZeJkZkH

---
_Generated by [Claude Code](https://claude.ai/code/session_01BLf35327oj8ArATZeJkZkH)_

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -116,6 +116,31 @@
             android:name="com.google.android.gms.car.application"
             android:resource="@xml/automotive_app_desc" />
 
+        <!-- FileProvider for Navidrome/Subsonic media-session cover art.
+
+             The platform media session (lock screen / Android Auto) loads
+             MediaItem.artUri in its OWN process, which can't read an app-private
+             file: path — so the cover Linthra fetched and cached is exposed to it
+             as a content:// URI served here instead. exported="false" +
+             grantUriPermissions: nothing is world-readable; the platform reads it
+             via the session, and audio_service decodes it in-process. It serves
+             ONLY the private cover cache (res/xml/media_artwork_paths.xml), whose
+             filenames are SHA-256 hashes of a credential-free reference — never a
+             username, token, salt, server URL, or auth query. No storage
+             permission is added. The authority is fixed (matches
+             kMediaArtworkAuthority in
+             lib/core/services/media_artwork_content_uri.dart) so it is identical
+             in debug and release. -->
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="io.github.thezupzup.linthra.mediaartwork"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/media_artwork_paths" />
+        </provider>
+
         <!-- Don't delete the meta-data below.
              This is used by the Flutter tool to generate GeneratedPluginRegistrant.java -->
         <meta-data
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -121,18 +121,20 @@
              The platform media session (lock screen / Android Auto) loads
              MediaItem.artUri in its OWN process, which can't read an app-private
              file: path — so the cover Linthra fetched and cached is exposed to it
-             as a content:// URI served here instead. exported="false" +
-             grantUriPermissions: nothing is world-readable; the platform reads it
-             via the session, and audio_service decodes it in-process. It serves
-             ONLY the private cover cache (res/xml/media_artwork_paths.xml), whose
-             filenames are SHA-256 hashes of a credential-free reference — never a
-             username, token, salt, server URL, or auth query. No storage
-             permission is added. The authority is fixed (matches
-             kMediaArtworkAuthority in
+             as a content:// URI served here instead. The MediaArtworkFileProvider
+             subclass grants the media consumers (Android Auto / SystemUI /
+             Bluetooth) read access to each cover URI when Linthra opens it, so
+             those processes can actually read it; exported="false" +
+             grantUriPermissions keeps nothing world-readable — only those
+             explicit per-URI read grants apply. It serves ONLY the private cover
+             cache (res/xml/media_artwork_paths.xml), whose filenames are SHA-256
+             hashes of a credential-free reference — never a username, token,
+             salt, server URL, or auth query. No storage permission is added. The
+             authority is fixed (matches kMediaArtworkAuthority in
              lib/core/services/media_artwork_content_uri.dart) so it is identical
              in debug and release. -->
         <provider
-            android:name="androidx.core.content.FileProvider"
+            android:name=".MediaArtworkFileProvider"
             android:authorities="io.github.thezupzup.linthra.mediaartwork"
             android:exported="false"
             android:grantUriPermissions="true">
--- a/android/app/src/main/kotlin/io/github/thezupzup/linthra/MediaArtworkFileProvider.kt
+++ b/android/app/src/main/kotlin/io/github/thezupzup/linthra/MediaArtworkFileProvider.kt
@@ -29,8 +29,10 @@ import java.io.FileNotFoundException
  * artwork, which are not served here.
  */
 class MediaArtworkFileProvider : FileProvider() {
+    // FileProvider.openFile returns a nullable ParcelFileDescriptor?, so the
+    // override must match (Kotlin rejects narrowing it to non-null).
     @Throws(FileNotFoundException::class)
-    override fun openFile(uri: Uri, mode: String): ParcelFileDescriptor {
+    override fun openFile(uri: Uri, mode: String): ParcelFileDescriptor? {
         grantReadToMediaConsumers(uri)
         return super.openFile(uri, mode)
     }
--- a/android/app/src/main/kotlin/io/github/thezupzup/linthra/MediaArtworkFileProvider.kt
+++ b/android/app/src/main/kotlin/io/github/thezupzup/linthra/MediaArtworkFileProvider.kt
@@ -1,0 +1,67 @@
+package io.github.thezupzup.linthra
+
+import android.content.Intent
+import android.net.Uri
+import android.os.ParcelFileDescriptor
+import androidx.core.content.FileProvider
+import java.io.FileNotFoundException
+
+/**
+ * FileProvider for Navidrome/Subsonic media-session cover art that also grants
+ * the platform media consumers temporary **read** access to each cover URI.
+ *
+ * Why this exists: the platform media session loads `MediaItem.artUri` in the
+ * consumer's *own* process (Android Auto's, SystemUI's for the lock screen,
+ * Bluetooth AVRCP), not Linthra's. The provider is `exported="false"`, so those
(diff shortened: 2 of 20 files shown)


TheZupZup:
**Status: parked — deferred from `v0.1.4` after a real-device Android Auto test. Do not merge as-is.**

Real-car test (Navidrome/Subsonic only): title/artist/progress/controls are correct, but the **cover stays a placeholder** in Android Auto and **playback felt less smooth**.

Root cause, traced through `audio_service` 0.18.18:
- A `file:` `artUri` **does** get its bitmap embedded into the session metadata in-process (`audio_service.dart:1058-1059` → `AudioService.java:787-791,809-813`), so a `content://`/FileProvider is **not** the missing piece. The real issue is **timing**: this PR fetches the cover asynchronously, so the bitmap lands ~100 ms–2 s *after* the track change in a second metadata push, and many head units snapshot the now-playing art at the track boundary and don't refresh it. (Jellyfin works because its `https://` `artUri` is present *synchronously* at the boundary.)
- The cover is requested at **original size** (no `getCoverArt?size=`), so the platform runs a synchronous full-resolution `BitmapFactory.decodeFile` on its main thread — slow / can OOM to a null bitmap, and a likely source of the smoothness regression. (Not a rebroadcast loop — it's exactly one extra media-item push per track, verified by code trace.)

Privacy is unaffected: `artUri` is only ever `file:` / token-free `http` / `null`; the cache filename is a SHA-256 of the credential-free reference; the authenticated URL is used once to fetch and never logged or persisted.

**`v0.1.4` ships the in-app cover fix only** — #170 plus the `null`-drop guard #171, both already in `main`.

Follow-up for a later release (requires real head-unit verification): request a **downscaled** cover, present the art **synchronously / pre-fetched** to beat the head-unit snapshot, and drop the async re-broadcast.

---

TheZupZup:
**Reworked** to address the real-device findings (still a draft, still **not for `v0.1.4`** — needs an on-device car test first).

The earlier cut fetched the cover asynchronously and at full size. Tracing `audio_service` 0.18.18 showed a `file:` artUri *does* get its bitmap embedded in-process, so the scheme wasn't the issue — the cover just arrived in a second metadata push *after* the track change (ignored by head units that snapshot the now-playing art), and the synchronous full-res decode was a likely jank/OOM source.

What changed:
- **Downscaled cover** — the media session now requests `getCoverArt?size=512`, so the download is small and the platform decode is cheap (`SubsonicEndpoints.coverArt` / `SubsonicArtwork.resolve` gained an optional `size`; the in-app full-size render from #170 is unchanged).
- **Pre-warm** — a new `MediaArtworkPrewarmService` (mirroring the stream preloader) warms the now-playing + next few covers into the cache off the playback path, so a cover is usually cached **before** its track reaches the now-playing card — letting the handler attach it **synchronously** and beat the snapshot.
- **Synchronous handler, no re-broadcast** — the handler now reads an already-cached `file:` via `MediaArtworkSource.cached(...)` while building the `MediaItem` and never triggers a fetch or re-broadcast on the playback path. A cover warmed mid-track is carried by the controller's next natural state tick.

Privacy unchanged: `artUri` is only ever a private `file:` / token-free `http` / `null`; cache filename is a SHA-256 of the credential-free reference; the (now downscaled) authenticated URL is used once and never logged/persisted; cast still omits Subsonic art.

Local checks: full suite **1689 green**, `flutter analyze` clean, `dart format` clean, secret-scan clean.

**Please re-test on your car when you can:** does the Subsonic now-playing cover now appear in Android Auto, and does playback feel smooth again? Head units vary in whether they refresh art mid-track, so on-device confirmation is the gate before this is considered done. If it works, it can target a later release (not the `v0.1.4` hotfix).

---

TheZupZup:
**Parked as a draft follow-up — not for `v0.1.4`** (per the second on-device test). `v0.1.4` stays the in-app Navidrome/Subsonic cover from #170.

### Second real head-unit test (Navidrome/Subsonic)
- ✅ **Playback smoothness regression is fixed** — the downscale + pre-warm + purely synchronous handler did the job.
- ❌ Android Auto **still shows a placeholder** for the cover.
- ✅ In-app Navidrome cover art still works.

### Why a private `file:` can't work for Android Auto (traced in `audio_service` 0.18.18)
Even though the cover is now cached and attached **synchronously** at the track change, Android Auto loads now-playing album art from the **art URI in its own process** (`METADATA_KEY_DISPLAY_ICON_URI`) — the Binder-friendly path Google steers media apps toward instead of shipping a large bitmap across processes. An app-private `file:///data/.../media_session_artwork/<hash>.img` simply **isn't readable from Android Auto's process**. Jellyfin's cover works only because its `artUri` is `https://`, which Android Auto can fetch itself. (`audio_service` does also decode a `file:` into an embedded bitmap in-process, but that path isn't what this head unit renders from.)

### Follow-up direction
Serve the **same cached cover** as a credential-free **`content://` URI via a `FileProvider`** and set *that* as `artUri`:
- Android Auto can read a granted `content://` (and `audio_service` also decodes it in-process), so it's the robust, recommended mechanism — it generalizes Jellyfin's "AA loads the URI" path to local covers.
- Privacy is preserved: the `content://` path is an opaque hashed filename — no username/salt/token/server URL, exactly like the current `file:` cache. The credential never reaches `artUri` or the metadata.
- Optionally pair with `androidArtDownscaleSize` so any embedded bitmap stays small enough to cross to the car.
- Requires a `FileProvider` (manifest + `file_paths.xml`) and read-grants to the bound media-browser client — and **another on-device car test**, which is why it's a separate, post-`v0.1.4` effort.

The downscale + pre-warm + synchronous-handler work on this branch is the right foundation for that; the follow-up only swaps the `file:` URI for a `content://` one. Findings are also recorded in `docs/android-auto.md` / `docs/providers.md`.

I'm leaving this parked and **not** watching its CI. Say the word when you'd like me to build the `content://` follow-up.

---

TheZupZup:
**`content://` FileProvider implemented** — this is now a `v0.1.4` candidate, gated on your car test.

### What changed since the parked version
- **FileProvider** (`androidx.core.content.FileProvider`, fixed authority `io.github.thezupzup.linthra.mediaartwork`, `exported=false`, no storage permission) serving **only** the private cover cache (`res/xml/media_artwork_paths.xml`).
- `MediaArtworkCache` now hands the session a **`content://…/media_artwork/<sha256>.img`** URI instead of `file:` — the platform/Android Auto process can read a `content://` (and `audio_service` decodes it in-process) but not an app-private `file:`. Everything else (server-downscaled fetch, hashed cache key, pre-warm, synchronous handler read off the playback path) is unchanged.
- **Downscaled the embedded album-art bitmap** (`artDownscaleWidth/Height = 256`) so the bitmap `audio_service` puts in the session metadata is small enough to survive cross-process delivery to the car. Artwork-only; no audio-playback change.

### Privacy (unchanged guarantees)
Catalog stays `subsonic-cover:<id>`; the authenticated `getCoverArt` URL is generated only inside Linthra, used once to fetch, and **never** persisted/logged/exposed or used as `artUri`. The `content://` path is `…/media_artwork/<sha256>.img` — no username/token/salt/server URL/auth query anywhere (URI, filename, metadata, logs, DB). Cast still omits Subsonic art; in-app (#170), Jellyfin, and local artwork untouched; a failed fetch → no art, never blocks playback.

### Gates
- ✅ Local: `flutter analyze` clean, `dart format` clean, **full suite 1692 green**, secret-scan green.
- ⏳ CI: Secret-scan ✅ already; **Flutter checks** + **Build debug APK** running (I'm watching — will fix any failure).
- ⏳ **The real one — your car:** does the Navidrome/Subsonic now-playing cover now show in Android Auto, and does playback stay smooth?

If the cover shows and playback is smooth, this is good for `v0.1.4`. If it's still blank on your head unit, the remaining lever is granting Android Auto explicit read access to the `content://` (native, more involved) — but the `content://` URI + downscaled bitmap together should do it. Staying a draft until you confirm in the car.

---

Issue #351, opened 2026-08-11
[Dart] Debounce Library search so huge catalogs don't work on every keypress


TheZupZup:
Small performance win for the Library search.

Right now the search updates immediately on every character typed. That's fine on a small library, but with tens or hundreds of thousands of tracks it means doing a bunch of filtering/folding work while the user is still typing.

What I want:
- add a short debounce around the Library search input, around 150–250 ms
- cancel the old pending search when the user keeps typing
- keep clearing the search feeling immediate
- add a widget test that proves we don't filter for every keystroke
- don't change which tracks/albums/artists actually match

Please keep this issue just about the debounce. Precomputed search keys / Rust integration can be separate work.

Likely areas:
- `lib/features/library/library_screen.dart`
- `lib/features/library/library_search.dart`
- matching Library widget tests

Expected size: small. Great first Flutter performance issue.

PR #357, opened 2026-08-12
fix: debounce library search

Borhan2004:


TheZupZup:
Thanks a lot for the PR! 😄 I really appreciate you taking the time to help Linthra.

I took a look and it’s already in really good shape. I only spotted one small edge case before we merge it:

If the user types something and taps **Clear before the 300 ms debounce finishes**, the pending timer can still fire afterward and put the old query back.

Could you cancel the pending debounce inside `_clearSearch()` too, and add a small test for:

type something → wait ~100 ms → clear → wait past 300 ms → the library should stay unfiltered

It would also be good to make sure a pending debounce can’t bring the query back after changing tabs.

Other than that, really nice work, and thanks again for jumping in on this issue! 😄

Borhan2004:
Thanks for catching this edge case. I've updated the PR to cancel pending debounce timers when clearing the search and when changing tabs, and added regression tests for both scenarios. All 8 debounce tests and 58 library tests are passing, with no analyzer issues.

TheZupZup:
Thanks! @Borhan2004 😄 I really appreciate you fixing those so quickly.
I think the PR may have been closed by accident though, it’s still showing as closed on my side. Could you reopen it and make sure the updated commit is pushed? Then I’ll take another look. Thanks again!

TheZupZup:
@Borhan2004 I’m going to reopen the PR so you can push the changes there.

Once the updated commit is up, I’ll take another look. Thanks again for fixing those edge cases!

Borhan2004:
Thanks! I’ve pushed the requested edge-case fixes. The pending debounce is now cancelled when clearing the search or switching tabs, with regression tests covering both scenarios. All checks are passing and there are no merge conflicts. Please take another look when you have a chance.

TheZupZup:
Fixes #351

PR #360, opened 2026-08-12
test: add Now Playing TalkBack tests

Borhan2004:
Adds focused accessibility tests for the Now Playing screen covering play/pause, next/previous, shuffle/repeat semantics, and the seek control.


TheZupZup:
Hey @Borhan2004 , thanks again for jumping on another issue 😄

I just noticed something interesting while looking at the CI failure: it looks like you're using a much newer Flutter version than Linthra is currently pinned to.

Linthra is still on Flutter 3.27.4, so some of the semantics/test APIs you're using simply don't exist in the version we're building with yet.

Before asking you to change your PR around an older SDK, I'm going to look at upgrading Linthra's Flutter version first. If the newer APIs make sense for the project, I'd rather move Linthra forward than make you work backwards.

Thanks again for the PR and for coming back with another contribution, I really appreciate it.

TheZupZup:
Everything is good Thank you very much @Borhan2004        Closes #334

PR #361, opened 2026-08-12
chore: upgrade Flutter toolchain to 3.44.7

TheZupZup:
Upgrade Linthra's Flutter toolchain from 3.27.4 to the current 3.44.7 stable hotfix line so the project can use and test against the modern Flutter semantics APIs that surfaced in #360.

This first pass updates the project pin and the PR/build workflows that exercise normal Flutter CI, debug APKs, sponsor simulation, sponsor APKs, and Drift generation.

Before merge I want this PR to prove the upgrade rather than just change a version number:
- lockfile-enforced `pub get`
- formatting
- `flutter analyze`
- full Flutter tests
- debug APK build + Android release lint
- sponsor simulation builds
- no F-Droid/reproducibility regression

I will keep the existing committed lockfile/reproducible build model intact. No release/tag/version bump is included here.

Related: #360


Borhan2004:
Thanks for taking this forward. I noticed the RepeatMode import collision and the related test fixes in the migration commits — that looks like it addresses the compatibility issue we ran into in #360.

I’ll keep #360 unchanged for now and rebase it onto main once this upgrade is merged. I’ll then rerun the TalkBack tests and the full Flutter checks against 3.44.7.

Borhan2004:
Just noticed the four CI checks are still failing. Hopefully they’re not too painful to sort out. 😄 It might be worth checking the first failing step in each workflow to see whether they’re related to the 3.44.7 migration or just specific to individual workflows. Maybe that can help narrow things down a bit. Happy to test #360 once the upgrade is ready!

Issue #348, opened 2026-08-11
[Dart] Make self-hosted server URLs work with IPv6 literals


TheZupZup:
Small networking one for self-hosters.

Our server URL normalizers rebuild URLs from `Uri.host`. IPv6 literal hosts need brackets in the final URL, so I want this covered properly instead of finding out from a user with an IPv6-only setup later.

What I need:
- add a failing test first with something like `https://[::1]:8096`
- make sure the normalized result stays parseable
- cover Jellyfin, Navidrome/Subsonic and Plex behaviour
- keep all the existing hostname/port/path tests green

If we land the shared URL-normalizer issue first, this can be fixed once there. If not, keep the provider behaviour consistent.

Please don't change normal URL behaviour while fixing IPv6.

Likely areas:
- `lib/core/sources/*_server_url.dart`
- matching tests under `test/core/sources/`

Expected size: tiny to small. Great first issue.

PR #479, opened 2026-08-18
Share one server URL normalizer across providers

### LINKED ISSUE #350: [Dart] Share one server URL normalizer across providers
Small cleanup that should make future networking fixes less annoying.

Jellyfin, Navidrome/Subsonic and Plex each have almost the same server URL normalization logic. Right now a URL fix can easily get changed in one provider and forgotten in the others.

What I want:
- extract the common trim / scheme / host / port / path normalization into one shared helper
- keep provider-specific error messages the same
- keep Subsonic's special REST suffix handling where it belongs
- keep all existing provider URL tests passing
- don't change user-visible behaviour as part of the refactor

A thin provider adapter around one shared implementation is probably enough.

Likely areas:
- `lib/core/sources/jellyfin_server_url.dart`
- `lib/core/sources/subsonic_server_url.dart`
- `lib/core/sources/plex_server_url.dart`
- a small new shared helper under `lib/core/sources/`

Expected size: small. Nice first refactor PR.

jpdexter101-lang:
Closes #350.

## What this does

Jellyfin, Plex, and Subsonic/Navidrome each had their own near-identical copy of the trim / default-scheme / parse / validate / rebuild logic for server addresses. Pulls the shared mechanics into `ServerUrlNormalizer.parse()` in `lib/core/sources/server_url_normalizer.dart`, returning a `ParsedServerUrl` (scheme/host/port/path plus a `toBase()` rebuilder) or throwing a `ServerUrlParseFailure` carrying an error kind.

Each provider's `normalize()` is now a thin adapter: call the shared parser, translate the failure kind into its own typed exception with its own message text, and build the final URL. Subsonic still does its own `/rest`-suffix stripping on top, unchanged, exactly where the issue asked for it to stay.

## What this doesn't do

No behavioural change. Every existing provider-specific error message is preserved verbatim, and `toBase()` rebuilds the URL exactly the way each provider's old inline code did — including the not-yet-fixed IPv6-literal-host bug, left alone here on purpose. That's a separate, focused follow-up PR (#348) stacked on this branch, since the issue itself notes IPv6 is easiest to fix once this lands.

## Testing

- All existing Jellyfin and Subsonic normalizer tests pass unchanged.
- Added `test/core/sources/plex/plex_server_url_test.dart` — Plex had no normalizer test coverage at all before this, so the refactor had nothing local to check it against for that provider. 9 new tests mirroring the existing Jellyfin/Subsonic style.
- `flutter analyze` clean on all four changed files.
- Full `flutter test` run against a clean checkout of `main` to confirm a handful of pre-existing failures elsewhere (an offline-cache resolver test, some CI/tooling guardrail tests) are unrelated to this change, not introduced by it.

### DIFF
--- a/lib/core/sources/jellyfin/jellyfin_server_url.dart
+++ b/lib/core/sources/jellyfin/jellyfin_server_url.dart
@@ -1,3 +1,4 @@
+import '../server_url_normalizer.dart';
 import 'jellyfin_exception.dart';
 
 /// Validates and normalizes the server address the user types into settings.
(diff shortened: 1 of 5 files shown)


TheZupZup:
Thank you so much for this! This was a really good catch. Having the same server URL logic duplicated between Jellyfin, Plex and Subsonic/Navidrome was definitely something that could become painful later. I really like that you cleaned this up without changing the actual behavior, and adding the missing Plex tests was a really nice touch too. Really appreciate the work on this!

PR #480, opened 2026-08-18
Make self-hosted server URLs work with IPv6 literals

### LINKED ISSUE #348: [Dart] Make self-hosted server URLs work with IPv6 literals

jpdexter101-lang:
Closes #348. Stacked on #479 (the shared-normalizer refactor) -- this is the fix the issue says becomes a single small change once that one lands, so please review/merge #479 first.

## What this fixes

`Uri.host` reports an IPv6 literal without its brackets (`::1`, not `[::1]`), so the normalizers were rebuilding an address as `scheme://::1:port`. That string is unparseable -- nothing can tell where the address ends and the port begins anymore. Confirmed empirically: `Uri.tryParse('https://::1:8096')` returns `null`. Anyone running Jellyfin/Plex/Navidrome on an IPv6-only server would hit this immediately.

## Testing (added first, confirmed failing pre-fix)

- One `https://[::1]:port` case per provider (Jellyfin, Plex, Subsonic)
- A full (non-loopback) IPv6 address with a subpath for Jellyfin
- A combined case for Subsonic: an IPv6 host plus a pasted `/rest` suffix, to make sure the two fixes compose correctly

All five fail against pre-fix code (verified before writing the fix), all five pass after it. Full existing suite stays green.

## Fix

`ParsedServerUrl.hostForUrl` brackets the host when it contains a colon -- the one thing that reliably marks an IPv6 literal, since a bare DNS hostname or IPv4 address never contains one. `toBase()` now writes that instead of the raw host. Because this lives in the shared normalizer from #479, one change fixes all three providers at once, which is exactly why the issue suggested landing the refactor first.

### DIFF
--- a/lib/core/sources/server_url_normalizer.dart
+++ b/lib/core/sources/server_url_normalizer.dart
@@ -37,12 +37,24 @@ class ParsedServerUrl {
   });
 
   final String scheme;
+
+  /// As reported by [Uri.host]. For an IPv6 literal this has no brackets
+  /// (e.g. `::1`, not `[::1]`) -- that's how Dart's own URI parsing
+  /// reports it. Use [hostForUrl], not this field directly, when
+  /// rebuilding a URL string.
   final String host;
+
   final int? port;
 
   /// Trailing slashes already trimmed; query and fragment already dropped
   /// ([Uri] parsing drops those from `.path` itself).
   final String path;
+
+  /// [host], bracketed when it's an IPv6 literal -- safe to drop straight
+  /// into a rebuilt URL string. An IPv6 literal is the only kind of host
+  /// [Uri.host] ever reports containing a colon (a bare DNS hostname or
+  /// IPv4 address never does), so that's a reliable and cheap check.
+  String get hostForUrl => host.contains(':') ? '[$host]' : host;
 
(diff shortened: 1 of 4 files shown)


TheZupZup:
Hi! @jpdexter101-lang  #479 is merged now, thank you so much.

Could you update/rebase #480 on the latest `main` and run `dart format` on `test/core/sources/jellyfin/jellyfin_server_url_test.dart`? The CI is only failing on formatting right now. Once that’s done I’ll recheck it and we should be good to merge. Thanks! 🙂

Issue #460, opened 2026-08-16
[Desktop] Improve accessibility labels and focus order


TheZupZup:
Part of #376.

## Goal
Audit desktop controls for meaningful semantics, predictable focus order and accessible actions.

## Likely code areas
- shared widgets/semantics wrappers
- `lib/features/library/`
- `lib/features/player/`
- desktop navigation/dialog/settings widgets
- accessibility widget tests

## Requirements
- Prioritize playback controls, sidebar navigation, track rows, dialogs and server-management surfaces.
- Provide labels/state/value semantics for icon-only controls and sliders.
- Ensure focus order follows the visual/logical layout and remains usable in multi-pane views.
- Do not duplicate visible text unnecessarily for screen readers.
- Keep semantics useful on Android too when the widget is shared.

## Acceptance criteria
- [ ] Major interactive desktop controls expose useful names/roles/states.
- [ ] Focus order is predictable with keyboard navigation.
- [ ] Disabled/selected/playing states are communicated appropriately.
- [ ] Icon-only playback controls are understandable without vision.
- [ ] Representative semantics/focus tests are added.

PR #481, opened 2026-08-18
Give the play/pause button a proper button role

jpdexter101-lang:
Part of #460, not the whole thing. Scoping this to just the transport row since that's what the issue lists first and it's a clean chunk to review on its own. Sidebar nav, track rows, dialogs, and server management are separate follow ups, didn't touch any of that here.

Shuffle, repeat, next, and previous all go through IconButton, so they get a button role and enabled state for free. Play/pause is hand built from Container and InkWell instead, for the brand gradient and glow, and Tooltip only ever gave it a label. A screen reader user got a name for the control but it never got announced as an actual button the way every other transport control does.

There's already a talkback test for this exact button, and it quietly leaves out isButton for Play/Pause while asserting it for everything else in the same file. Strengthened it first to check for isButton/isEnabled/hasEnabledState the same way the other buttons already do, confirmed that fails against the unfixed widget with exactly those three flags missing, then fixed it by wrapping the button in Semantics(button: true, enabled: onTap != null), which merges straight into the semantics node Tooltip already creates.

flutter analyze and dart format both clean. Ran the full player test directory against this branch and against a clean checkout of main, same 3 unrelated failures in playback_resolution_test.dart and source_routing_test.dart either way, nothing new from this change.

### DIFF
--- a/lib/features/player/widgets/playback_controls.dart
+++ b/lib/features/player/widgets/playback_controls.dart
@@ -151,35 +151,44 @@ class _PlayPauseButton extends StatelessWidget {
 
     return Tooltip(
       message: playing ? 'Pause' : 'Play',
-      child: Container(
-        width: 72,
-        height: 72,
-        decoration: BoxDecoration(
-          shape: BoxShape.circle,
-          gradient: LinearGradient(
-            begin: Alignment.topLeft,
-            end: Alignment.bottomRight,
-            colors: <Color>[gradientTop, gradientBottom],
-          ),
-          boxShadow: <BoxShadow>[
-            BoxShadow(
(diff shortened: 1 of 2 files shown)


TheZupZup:
Really nice PR, thank you again! 😄 I especially like how focused this is — fixing just the Play/Pause semantics instead of trying to solve all of #460 in one go makes it super easy to review.

Good catch on the fact that Tooltip gave the control a label but not the actual button role like the other transport controls already have. Adding `Semantics(button: true, enabled: onTap != null)` is a clean fix, and strengthening the TalkBack test at the same time makes this much more solid.

Also really appreciate that you checked the existing player test failures against `main` instead of assuming they came from this change. Thanks for another clean contribution!

PR #483, opened 2026-08-18
Index tracks.source_id so a source re-sync doesn't full-scan the catalog

### LINKED ISSUE #349: [Dart/SQL] Index tracks.source_id so sync doesn't full-scan the catalog
Small database/performance issue.

Linthra deletes tracks by `source_id` when a source sync is replaced, but the shipped Drift table doesn't currently have an index for that lookup. On a big catalog that means scanning the whole tracks table just to remove one source.

What I want:
- add an index for `source_id` in the Drift schema
- bump the database schema version cleanly
- add the migration for existing installs
- add a migration/regression test so upgrades don't lose data
- regenerate Drift code like the project normally does

Please keep this focused on the access pattern we actually use. No random extra indexes just because they might be useful someday.

Likely areas:
- `lib/data/database/tables/tracks_table.dart`
- `lib/data/database/linthra_database.dart`
- `test/data/database/`

Expected size: small. Good first issue if you're comfortable with Drift migrations.

jpdexter101-lang:
Closes #349.

A source re-sync deletes that source's rows first (_deleteSource in drift_music_library_repository.dart, DELETE FROM tracks WHERE source_id = ?), but nothing narrowed that lookup, so every sync scanned the whole tracks table just to find one source's rows. Checked this directly rather than assuming: the delete's own query plan now reads SEARCH tracks USING INDEX tracks_source_id instead of a scan.

Added the index via @TableIndex on the Tracks table, bumped schema to v4, added the migration for existing installs. One thing worth calling out: createTable (used by the old v1 to v2 rebuild) only ever issues CREATE TABLE, never the indexes declared on a table, confirmed that by reading Migrator's source rather than assuming it. So every upgrade path needs the index added as its own explicit step no matter which version someone's coming from, only createAll (the fresh install path) already includes it for free.

Migration tests build real pre-migration tables by hand and run the actual MigrationStrategy over them, same convention the existing v2 to v3 tests already use. They confirm existing rows and every value survive untouched, the index exists and is actually used by the query planner afterward, and a v1 database migrating straight through to v4 in one pass still picks up the index, not just a v3 database. The one pre-existing test that asserted the post-migration version now checks against db.schemaVersion instead of a hardcoded 3, so it doesn't go stale the next time schemaVersion bumps again.

flutter analyze clean, dart format clean, full test/data/database suite plus drift_music_library_repository/drift_repository_wiring/track_mapper all pass with no regressions. Also built a debug APK from this branch and ran it on a physical phone to confirm the app installs and runs fine with the new schema, not just in the test harness.

### DIFF
--- a/lib/data/database/linthra_database.dart
+++ b/lib/data/database/linthra_database.dart
@@ -23,6 +23,10 @@ part 'linthra_database.g.dart';
 ///    one album (`library_grouping.dart`). Purely additive: existing rows keep
 ///    every value and read back with the new columns `null` until the next
 ///    source re-scan populates them.
+///  * **v4** — index on `tracks.source_id` (see `tracks_table.dart`), so a
+///    source re-sync's `DELETE ... WHERE source_id = ?` no longer scans the
+///    whole catalog to find the one source's rows. Purely additive: no rows
+///    or columns change.
 @DriftDatabase(tables: [Tracks])
 class LinthraDatabase extends _$LinthraDatabase {
   LinthraDatabase() : super(_openConnection());
@@ -32,7 +36,7 @@ class LinthraDatabase extends _$LinthraDatabase {
   LinthraDatabase.forTesting(super.executor);
 
   @override
-  int get schemaVersion => 3;
+  int get schemaVersion => 4;
 
   @override
   MigrationStrategy get migration => MigrationStrategy(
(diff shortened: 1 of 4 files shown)


jpdexter101-lang:
Curious if audiobook support is something you’ve ever thought about, specifically Audiobookshelf. I listen to a lot of audiobooks and there aren’t many good apps that hook into a self-hosted server and handle stuff like chapters and resume position properly.

No idea if that’s even something you want Linthra to do, just figured I’d ask before opening an issue for something you’ve already decided against. If it sounds interesting I’d probably mess around with it and see what I can get working. If not, no big deal.

TheZupZup:
@jpdexter101-lang Thanks again for this PR, really nice work on it. I especially appreciate that you checked the actual query plan and migration paths instead of just adding the index and calling it done.

And yes, you have the green light from me for audiobook support, especially Audiobookshelf. I think chapters + proper resume position with a self-hosted server could fit Linthra really well.

Feel free to open an issue for it and mess around with an implementation. I’d be happy to see where you can take it 🙂

Issue #484, opened 2026-08-18
Audiobookshelf support (chapters, resume position, self-hosted alternative to paid apps)


jpdexter101-lang:
I use audiobooks a lot and right now the only good option for listening from a self-hosted server is Bookcamp, and it only talks to Plex, and it's a paid subscription (~$1-2/month) for things like cross-device resume position and offline downloads. I'd love to see that same experience built into Linthra instead, backed by Audiobookshelf specifically, free and fully self-hosted end to end.

I don't want this to land as one giant PR. Same as the Linux port, I'd rather it come in as a few small, reviewable pieces.

**What I want**
- Audiobookshelf as a real provider (connection, auth, library browsing)
- Chapters as a first-class concept, not just long tracks, with a player that works just as well when a file has none
- Multi-file books play straight through automatically, no tapping "next" between files
- Real resume position that syncs across devices
- Adjustable playback speed that also follows you across devices, not just a local toggle that resets per phone
- Bookmarks, separate from resume position, since a listener deliberately marking a spot is a different thing from the app remembering where they stopped
- Sleep timer, chapter navigation
- Offline downloads

**Suggested PRs**

1. **Connection + library browsing** — server URL, auth, HTTP client, list the library. No playback yet.
2. **Domain model + chapters + local playback** — add chapters to the data model, get a book playing with chapter-aware seeking. Multi-file books load as one continuous queue so playback auto-advances between files with no manual tap, same as the existing queue behavior for music. Resume position local-only for now.
3. **Cross-device sync** — resume position and playback speed, both synced as Audiobookshelf's own server-side user settings, not just local state.
4. **Audiobook player UI** — speed control, chapter list/jump when available, always backed by a time scrubber and skip buttons so a chapterless file still has a real player, since plenty of real-world files (especially single-file torrent rips) have no embedded chapter metadata at all and Audiobookshelf's own scanner can't build chapters from just one unmarked file.
5. **Bookmarks** — Audiobookshelf already has real bookmark endpoints server-side, separate from resume position.
6. **Offline downloads** — may be able to lean on existing offline-cache infra, worth checking once in the code.

**Important**
- Don't touch the existing music providers or their UI
- Keep this behind its own clear seam, same as the Jellyfin/Subsonic/Plex providers are
- First milestone is just #1, working and reviewable on its own

TheZupZup:
@jpdexter101-lang This is awesome, thank you for taking the time to break it down like this.

[...]

PR #485, opened 2026-08-18
Audiobookshelf: connection, auth, and library listing (milestone 1 of #484)

jpdexter101-lang:
First slice of #484. Just the connection layer, no playback, no chapters, no domain model changes, exactly milestone 1 from the plan.

Mirrors the shape of the existing Jellyfin/Plex/Subsonic providers: a session model, a typed exception, a server URL validator built on the shared normalizer from #479 (so a future fix there covers this provider too), an endpoints file, an HTTP client behind an interface, and an authenticator coordinating testConnection/signIn.

Didn't want to guess at the request shapes from docs, so I actually went and read Audiobookshelf's own server source before writing any client code. Two things worth flagging since they're easy to get wrong otherwise:

POST /login needs an x-return-tokens: true header or the refresh token only comes back as an httpOnly cookie, which is useless for a client with no cookie jar. Found that by reading Auth.js's handleLoginSuccess directly.

GET /status includes app: "audiobookshelf" in the response, so fetchServerStatus uses that to confirm the address is genuinely an Audiobookshelf server before any credentials go out, same role Jellyfin's product-name check plays for that provider.

38 new tests (server URL validation, the HTTP client against realistic canned JSON, the authenticator's coordination logic through a fake client, session model round-tripping), all green. flutter analyze and dart format both clean.

### DIFF
--- a/lib/core/models/audiobookshelf_session.dart
+++ b/lib/core/models/audiobookshelf_session.dart
@@ -83,23 +83,24 @@ class AudiobookshelfSession {
       };
 
   /// Rebuilds a session from [toJson] output, or returns `null` if any
-  /// required field is missing/blank (e.g. a partially written or corrupted
-  /// record), so the app treats it as "not signed in" rather than crashing.
+  /// required field is missing, blank, or the wrong type (e.g. a partially
+  /// written or corrupted record), so the app treats it as "not signed in"
+  /// rather than crashing on a bad cast.
   static AudiobookshelfSession? fromJson(Map<String, dynamic> json) {
-    final String? baseUrl = json['baseUrl'] as String?;
-    final String? userId = json['userId'] as String?;
-    final String? accessToken = json['accessToken'] as String?;
-    if (baseUrl == null || baseUrl.isEmpty) return null;
-    if (userId == null || userId.isEmpty) return null;
-    if (accessToken == null || accessToken.isEmpty) return null;
+    final String? baseUrl = _string(json['baseUrl']);
+    final String? userId = _string(json['userId']);
+    final String? accessToken = _string(json['accessToken']);
+    if (baseUrl == null || userId == null || accessToken == null) {
+      return null;
+    }
     return AudiobookshelfSession(
       baseUrl: baseUrl,
       userId: userId,
       accessToken: accessToken,
-      refreshToken: json['refreshToken'] as String?,
-      userName: json['userName'] as String?,
-      defaultLibraryId: json['defaultLibraryId'] as String?,
-      serverVersion: json['serverVersion'] as String?,
+      refreshToken: _string(json['refreshToken']),
+      userName: _string(json['userName']),
+      defaultLibraryId: _string(json['defaultLibraryId']),
+      serverVersion: _string(json['serverVersion']),
     );
   }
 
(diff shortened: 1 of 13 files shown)


TheZupZup:
Thank you so much for picking this up and especially for taking the time to verify the Audiobookshelf server behavior directly instead of guessing from the API docs.

The x-return-tokens detail and the /status validation are exactly the kind of things that could have turned into really annoying bugs later, so I really appreciate the care you put into this.

I also really like that you kept this scoped to milestone 1 instead of trying to implement the entire provider in one huge PR. The 38 tests are a very nice bonus too.

I’m going to review the changes and CI carefully. Thanks again for contributing this I’m genuinely excited to see Audiobookshelf support starting to land in Linthra!

TheZupZup:
@jpdexter101-lang Thank you again for this PR, the structure is really clean and I like the way you kept milestone 1 focused.

I found 3 things I’d like to get tightened up before I merge it:

1. Can you re-check the real current Audiobookshelf `/status` response against the server version we want to support? Right now the client requires `app: "audiobookshelf"` plus `serverVersion`. I’m worried that may reject real current servers depending on the response shape. I’d rather validate against what the current server actually returns than rely on a field that may not be there everywhere.

2. This one is the important one: if the `/status` check fails, `signIn()` currently swallows that and still calls `/login`. I don’t want Linthra sending a username/password to an address we failed to validate first. Please make sure credentials are only sent after the same normalized base URL has successfully been confirmed as an Audiobookshelf server.

3. Small hardening thing: `AudiobookshelfSession.fromJson()` says corrupt persisted data should return `null`, but the direct `as String?` casts can still throw if a field has the wrong type. Can you make that fail safe too and add a regression test for wrong-typed persisted values?

None of these change the direction of the PR, I really like what you built here. I just want to make the connection/auth layer solid now because everything else in #484 is going to sit on top of it. Thanks again for putting this together.

jpdexter101-lang:
Good catches, thanks for the thorough review. Fixed all three:

1. Re-checked `/status` against the real handler across release tags, not just the current one. Turns out `app` and `serverVersion` were only added in v2.6.0, so anyone on v2.5.0 or earlier would've been rejected outright. Fixed the check to fall back to `isInit` (present in every version I checked back to v2.2.0) when `app` is absent, and made `serverVersion` nullable since older servers genuinely don't send it. Added a regression test for a pre-v2.6.0 response shape.

2. You're right, that was a real bug. `signIn()` now propagates a failed status check instead of swallowing it, so credentials only go out once the address is confirmed. Updated the test that was actually asserting the old (wrong) behavior.

3. Added the same safe-cast helper the API layer already uses to `AudiobookshelfSession.fromJson()`, plus a regression test for wrong-typed persisted fields.

Should have caught all three of these myself. Juggling a few projects at once and rushed the verification on this one. Appreciate you catching it before merge instead of after.

TheZupZup:
@jpdexter101-lang I really like the work you’ve been doing on Linthra. The way you handled the feedback here and came back with clean fixes makes me even more confident in this PR.

Everything looks good on my side now and CI is green, so I’m going to merge this one. Thanks again for putting so much care into it — I’m really happy to see Audiobookshelf support starting this strong.

PR #490, opened 2026-08-20
fix: harden Linux playback across system suspend/resume

Borhan2004:
## Summary
- Closes #466: Linux playback recovers predictably after system suspend/resume.
- Arms recovery only on `AppLifecycleState.paused` (not brief `inactive` dialogs); on Linux, a track that was **playing** is re-resolved/reloaded once on the same engine after a short backoff.
- User-paused tracks stay paused; failed recovery keeps the queue and surfaces the existing error + Retry UI; Android screen-on stays inert.
- Documents a manual suspend/resume matrix in `docs/linux-desktop.md`.

## Test plan
- [x] `flutter test` on linux suspend/resume recovery, Linux playback controller, ActivePlaybackController lifecycle, app playback lifecycle, unsupported controller, focus tests
- [x] `dart analyze` clean on changed Dart paths
- [ ] Manual matrix in `docs/linux-desktop.md`: sleep while playing local/remote, while paused, with Bluetooth/network delayed after wake, and 3× repeat cycles

PR #491, opened 2026-08-20
feat(linux): crash-safe playback state restoration

### LINKED ISSUE #465: [Linux] Add crash-safe playback state restoration
Part of #376.

## Goal
Restore useful non-secret playback state after an unexpected Linux app crash/restart without surprising auto-play.

## Likely code areas
- playback/queue state models under `lib/features/player/`
- persisted app/database/preferences state
- logical-track resolution
- startup restore logic/tests

## Requirements
- Persist only logical track identity, queue/modes and position where safe/useful.
- Never persist authenticated stream URLs, temporary provider tokens or raw resolver output.
- On restart, restore a paused/resumable state by default rather than automatically blasting audio.
- Re-resolve remote tracks through normal provider/session logic before playback.
- Handle missing local files/provider sign-out/version changes by dropping only invalid restore data.

## Acceptance criteria
- [ ] Unexpected restart can restore the logical queue/current item and a reasonable position.
- [ ] Remote playback uses a freshly resolved URL.
- [ ] App does not auto-play unexpectedly after crash/startup.
- [ ] Invalid/stale restore records fail safely without blocking app startup.
- [ ] Persistence/security/migration cases have tests.

Borhan2004:
## Summary
- Closes #465: after an unexpected Linux crash/restart, restore the logical queue, current item, modes, and position as a **paused** session (never autoplay).
- Persist only logical track identity (`jellyfin:` / `subsonic:` / `plex:` / local paths) — never authenticated stream URLs, tokens, or resolver output; remotes re-resolve through the normal provider path on play.
- Invalid/stale rows (missing local files, signed-out providers, corrupt/wrong-version records) are dropped safely and never block startup. Android behaviour is unchanged.

### DIFF
--- a/docs/linux-desktop.md
+++ b/docs/linux-desktop.md
@@ -246,6 +246,21 @@ widget.
 alone. If someone re-runs `flutter create --platforms=linux .` anyway, the
 checker above is what catches the reverted title and the lost SQLite seam.
 
+## Crash-safe playback restore
+
+On Linux, Linthra persists a small **logical** playback session while a track is
+queued — provider-namespaced track ids (`jellyfin:…`), local paths, shuffle/
+repeat modes, and position. It never writes authenticated stream URLs or
+provider tokens. After an unexpected process exit, the next launch restores that
+queue as **paused** (never autoplay); pressing play re-resolves remote tracks
+through the normal signed-in provider path. Missing local files, signed-out
+providers, and wrong-version/corrupt records drop only the invalid rows (or the
+whole session) and never block startup.
+
+Automated coverage: `test/core/models/persisted_playback_session_test.dart`,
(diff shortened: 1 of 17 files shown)


TheZupZup:
Hey @Borhan2004! Looks like merging #490 caused a conflict with this one 😅

Could you update the branch with the latest main and resolve it while keeping the changes from #490 intact?

Ping me when it’s ready and I’ll take another look 👀

Borhan2004:
Okay let me see

Borhan2004:
Updated the branch with latest `main` and resolved the `docs/linux-desktop.md` conflict.

Kept both:
- the **Suspend / resume** matrix + limitations table row from #490
- this PR’s **Crash-safe playback restore** section

Ready for another look when you have a moment.

PR #509, opened 2026-08-25
test: keep Android audio formats in sync

### LINKED ISSUE #352: [Kotlin/Dart] Keep supported local audio formats in sync
Small cross-language cleanup.

The supported local audio formats are listed in more than one place right now: Dart has its list/mappings and the Android SAF scanner has another copy. That's easy to forget when someone adds support for a new format later.

What I want:
- make it hard for the Kotlin and Dart format lists to drift apart
- either make the Android scanner genuinely permissive and let Dart be the final source of truth, or add a simple regression check that proves both sides agree
- fix any comments that currently make the ownership sound clearer than it really is
- keep the current supported formats working exactly the same

Please don't add new formats as part of this issue unless it's needed for a test. This is about one source of truth / a guardrail.

Likely areas:
- `lib/core/sources/local/audio_file_types.dart`
- `lib/core/sources/audio_file_extension.dart`
- `android/app/src/main/kotlin/.../SafDocumentScanner.kt`
- focused tests

Expected size: small. Nice first issue if you like Kotlin + Dart together.

jpdexter101-lang:
## What this fixes

Fixes #352.

The local audio formats are defined in both Dart (`AudioFileTypes`) and the Android SAF scanner (`SafDocumentScanner`). Those lists currently match, but a future format addition could update one side and silently make Android folder scans disagree with the rest of the app.

This adds a focused regression test that reads the Android scanner's extension list and asserts exact set equality with the Dart source of truth. It catches additions, removals, and spelling drift without changing the currently supported formats or runtime behavior.

## Verification

- `dart format test/core/sources/local/audio_file_types_test.dart`
- `flutter test test/core/sources/local/audio_file_types_test.dart` — 13 tests passed
- `git diff --check` — passed

### DIFF
--- a/test/core/sources/local/audio_file_types_test.dart
+++ b/test/core/sources/local/audio_file_types_test.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:flutter_test/flutter_test.dart';
 import 'package:linthra/core/sources/local/audio_file_types.dart';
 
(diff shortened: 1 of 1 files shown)


jpdexter101-lang:
Closing this attempt because the Android Kotlin boundary triggers the repository owner approval security gate. The parity test is valid, but I am switching to a Dart only accessibility issue better suited to a quick external contribution.

PR #510, opened 2026-08-25
fix(ui): keep mini-player usable with large text

### LINKED ISSUE #340: [Dart] Keep the mini-player clean with large text
Small UI one for someone who wants an easy Flutter PR.

I want the mini-player to stay clean even when Android font size is cranked up. Right now it looks good at normal size, but this is the kind of thing that can get ugly fast on accessibility settings if we don't test it.

What I need:
- add widget coverage with larger text scaling
- make sure title / artist / source don't overflow the 64dp mini-player
- keep the important controls reachable
- use ellipsis or a small layout adjustment if needed

Please don't redesign the whole mini-player in this issue. I just want it to survive large text properly.

Likely areas:
- `lib/features/player/mini_player.dart`
- `test/features/player/`

Expected size: small. Nice first Flutter issue.

jpdexter101-lang:
## What this fixes

Fixes #340.

The mini-player has a fixed 64dp height. With Android's larger text settings, the title and metadata can outgrow that bar and make the transport control unreachable or produce a layout overflow.

The bar now keeps its compact 64dp height at normal text size, but grows proportionally when the user's text scaler requires more vertical room. The existing ellipsis behavior and controls remain unchanged.

## Verification

- `dart format lib/features/player/mini_player.dart test/features/player/mini_player_test.dart`
- Added widget coverage at `TextScaler.linear(2.5)` checking that no exception occurs and the Pause control remains present.
- `flutter test test/features/player/mini_player_test.dart` was attempted locally but could not compile because the Windows Pub cache is missing `file_picker-8.3.7/lib/file_picker.dart`; this is an environment dependency-cache issue unrelated to the changed files.
- `git diff --check` — passed

### DIFF
--- a/lib/features/player/mini_player.dart
+++ b/lib/features/player/mini_player.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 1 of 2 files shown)


TheZupZup:
Nice fix, clean and focused. Thanks for covering the large text case too ❤️

Issue #556, opened 2026-09-04
[Desktop] Polish adaptive Linux layouts for a native music-player feel


TheZupZup:
Part of #376.

## Goal
Polish Linthra's feature layouts at desktop widths so the Linux app feels intentionally designed for mouse/keyboard and large windows rather than looking like Android screens stretched across a monitor.

This builds on the desktop shell/navigation work. Keep the same Linthra product identity and shared feature logic; adapt presentation density, spacing and panel behavior for desktop.

## Current problem
Even with a desktop shell, individual library/player views can still feel mobile-first if cards, controls and content remain sized and arranged for a narrow handset. Wide Linux windows should make useful use of available space without becoming visually sparse or oversized.

## Design direction
Use responsive layouts that progressively take advantage of width. Examples include:
- denser album/artist grids with sensible maximum card sizes;
- constrained readable content widths instead of stretching text edge-to-edge;
- multi-column/pane layouts where existing content naturally supports it;
- persistent or side-by-side Now Playing / queue / lyrics regions at sufficiently wide widths where this can be done using existing features;
- desktop-appropriate headers/toolbars and action placement;
- pointer hover/focus feedback where existing controls benefit from it;
- sensible spacing and hit targets for mouse usage without breaking touch support.

These are examples, not permission to invent new product features.

## Likely code areas
- library/album/artist/song presentation widgets
- Now Playing / queue / lyrics presentation
- responsive layout helpers and breakpoints
- shared cards/lists/toolbars used by Linux desktop views
- widget/golden tests where practical

## Requirements
- Preserve existing business logic, providers and route models.
- Prefer responsive composition over `Platform.isLinux` branches scattered through feature widgets.
- Keep Android/mobile layouts intact at narrow widths.
- Avoid fixed pixel layouts that only look correct on one monitor.
- Make album/artist grids scale by available width while preserving useful card dimensions.
- Avoid giant mobile-style controls simply expanding into empty desktop space.
- Where a wide view can display two useful existing surfaces together (for example content + queue/Now Playing), use shared state rather than duplicate controllers.
- Maintain accessibility semantics, keyboard focus and touch compatibility.
- Ensure scroll behavior remains natural with mouse wheel/trackpad.
- Avoid horizontal overflow at common desktop scaling factors.
- Keep theme/color/branding consistent with Linthra rather than copying another music player.
- Do not add new Flatpak permissions or Linux-native integrations in this issue.
- Do not regress Android, F-Droid or Google Play builds.

[...]

Issue #555, opened 2026-09-04
[Desktop] Add a real Linux desktop shell with persistent navigation


TheZupZup:
Part of #376.

## Goal
Give Linthra a real desktop application shell on Linux instead of presenting the Android/mobile navigation model at desktop widths.

The Linux app should keep Linthra's existing visual identity and shared feature screens, but desktop-sized windows need persistent navigation and clearer structural regions that feel intentional on GNOME/KDE.

This issue is about the desktop shell/navigation structure. Do not turn it into a full visual redesign of every screen.

## Current problem
Linthra can run as a native Linux app, but on a wide desktop window the experience still feels too close to the Android layout scaled up. Desktop users should not need to navigate through a phone-shaped shell when there is enough width for persistent navigation.

## Design direction
At desktop widths, introduce a proper shell with a persistent left navigation region/sidebar for the main destinations. The exact Flutter widget implementation is up to the PR, but the behavior should resemble a serious desktop music player rather than a stretched handset UI.

Likely primary destinations include the routes Linthra already exposes, such as:
- Home / library landing view;
- Songs;
- Albums;
- Artists;
- Playlists where available;
- configured music sources/providers;
- Settings.

Do not invent new product features just to fill the sidebar. Reuse the existing navigation model and routes.

## Likely code areas
- app/root navigation shell
- shared route/navigation state
- responsive layout helpers/breakpoints
- desktop-specific presentation widgets under the existing platform seams
- widget/golden tests where practical

## Requirements
- Preserve one shared Linthra application, not separate Android and Linux codebases.
- Keep Android/mobile navigation behavior intact at mobile widths.
- At desktop widths, replace phone-oriented bottom navigation/drawer dependence with persistent desktop navigation where appropriate.
- Navigation selection must remain synchronized with the existing route/state model.
- Deep links/back navigation must continue to work.
- The shell must resize cleanly rather than assuming one fixed monitor size.
- Support common Linux desktop window sizes, including narrower windows where the shell must degrade gracefully.
- Keep provider/library/player business logic out of desktop presentation widgets.
- Keep existing theming and Linthra branding; this issue is structural, not a rebrand.
- Do not add platform checks directly throughout feature screens. Keep desktop behavior behind reusable responsive/platform presentation seams.
- Do not regress Android, F-Droid or Google Play builds.

## Desktop behavior to validate
- 1280×720 class window;
- 1440p/ultrawide class window;
- resizing from wide to narrow while the app is running;
[...]

PR #549, opened 2026-09-03
ci(flatpak): build and launch packaging on clean runners (#444 #445)

TheZupZup:
Closes #444.
Closes #445.

## Goal

Add a dedicated, secret-free Flatpak CI path that proves packaging-relevant changes can build the real generated Flatpak on a clean GitHub-hosted runner, install that exact package, and reach a real Linthra desktop window before the change reaches Flathub.

## What this PR adds

- `.github/workflows/flatpak-build.yml`
  - runs on packaging-relevant PR/push changes plus manual dispatch;
  - keeps repository permissions read-only and uses no secrets;
  - validates Linthra's committed packaging metadata;
  - installs Flatpak/flatpak-builder/AppStream/desktop/headless-X11 validators;
  - resolves SDK/runtime prerequisites from the generated manifest via Flathub;
  - builds the real generated manifest with module-cache reuse disabled;
  - exports and verifies a local Flatpak repository;
  - installs Linthra from that repository and runs the packaged app in Xvfb;
  - requires a real `Linthra` X11 window to appear, then terminates and cleans up;
  - does not use `actions/cache`, so correctness cannot depend on a warm cache.
- `scripts/flatpak_launch_smoke.sh`
  - installs only from the local unsigned CI repository;
  - launches the real Flatpak app ID inside an isolated Xvfb/D-Bus session;
  - fails if the process exits early or no Linthra window appears within the timeout;
  - removes the temporary installation and local remote on exit.
- `test/tooling/flatpak_ci_guardrails_test.dart` and `test/tooling/flatpak_launch_smoke_test.dart`
  - lock down the read-only/fork-safe workflow and launch-smoke contracts.
- `docs/flatpak-ci.md`
  - documents exactly what CI proves and the equivalent local commands.

## Scope

This PR does not change runtime Flatpak permissions, provider/application behavior, Android behavior, dependency versions, the generated manifest, or Flathub submission metadata.

## Validation strategy

The PR itself is the first real clean-runner build/install/launch execution of this workflow. If GitHub-hosted Ubuntu exposes a runner-specific Flatpak requirement, fix it here rather than weakening the sandbox/build proof.

The stricter offline-source smoke (#442) and installed audio/local-library smokes (#446/#447) remain separate.

No merge is requested until all repository checks and the new Flatpak build/install/launch job are green.

PR #557, opened 2026-09-04
fix(linux): make the running window identify as io.github.thezupzup.linthra (#554)

### LINKED ISSUE #554: [Linux] Fix Flatpak/Wayland application identity and dock icon
Part of #376.

## Goal
Make Linthra identify itself consistently as `io.github.thezupzup.linthra` on Linux so GNOME/KDE/Wayland show the real Linthra icon and application identity instead of a generic Wayland/default window icon.

This issue is about desktop identity only. Do not redesign the UI here.

## Current problem
The Flatpak installs Linthra's desktop entry and scalable icon, but the running Linux window can still appear with a generic Wayland/default icon or fail to group correctly with the installed launcher.

The installed package identity, desktop file, icon name, Flutter/GTK window identity and Wayland/X11 application identity must all agree.

## Likely code areas
- `linux/runner/`
- `linux/CMakeLists.txt`
- `linux/packaging/io.github.thezupzup.linthra.desktop`
- Flatpak manifest/install rules
- `tool/branding/linthra_icon.svg`
- Linux runner validation tooling/tests

## Requirements
- Keep `io.github.thezupzup.linthra` as the canonical application ID.
- Investigate the actual GTK/Flutter Linux runner window identity instead of papering over the issue only in the `.desktop` file.
- Ensure GNOME Wayland associates the running window with the installed Linthra desktop entry.
- Ensure KDE Plasma/Wayland also shows the Linthra icon and groups the window correctly.
- Keep X11 fallback behavior correct where applicable.
- Do not introduce duplicate launcher entries or a second application ID.
- Reuse the existing official Linthra icon; do not create a new logo for Linux.
- Keep Android/F-Droid/Google Play application identity untouched.
- Preserve Flatpak sandbox restrictions.

## Validation
Test at minimum:
- launch from the application menu;
- launch with `flatpak run io.github.thezupzup.linthra`;
- GNOME Wayland dock/overview identity;
- KDE Plasma Wayland task manager identity;
- title/window grouping after closing and reopening the app.

## Acceptance criteria
- [ ] GNOME Wayland shows the Linthra icon for the running Flatpak instead of a generic Wayland/default icon.
- [ ] KDE Plasma Wayland shows the same Linthra icon and correct application grouping.
- [ ] Launcher, running window, desktop file, icon and Flatpak all use `io.github.thezupzup.linthra` consistently.
- [ ] `flatpak run io.github.thezupzup.linthra` and launcher startup behave the same.
- [ ] X11 fallback remains functional.
- [ ] Existing Linux/Flatpak validation tests are extended where practical to prevent identity drift.
- [ ] Android packaging/release behavior is unchanged.

## Scope guard
Do not redesign navigation, Now Playing, album grids or other desktop UI in this issue. Keep this PR focused on Linux application/window identity.

TheZupZup:
Closes #554.

The Flatpak already installed the desktop entry, the icon and the AppStream component under `io.github.thezupzup.linthra`. What did not fully agree with them was the running GTK window, so a launcher could show a generic icon or fail to group the window with its own entry.

I went looking at the actual GTK/GDK identity rather than patching the `.desktop` file, and found three separate gaps. Everything below is verified against GTK 3.24 sources and reproduced with a minimal GtkApplication built the same way the runner is, run under Xvfb with the icon installed in a hicolor theme.

## What GTK/Wayland/X11 was seeing before

**Wayland was already correct.** GTK 3 sends `xdg_toplevel.set_app_id()` straight from `g_get_prgname()` (`gdk/wayland/gdkwindow-wayland.c`, `gdk_wayland_window_create_xdg_toplevel`), and the runner already calls `g_set_prgname(APPLICATION_ID)`. So the app id GNOME and KDE match against under Wayland was right.

**X11 was not.** `gtk_window_init()` captures the WM_CLASS pair as `g_get_prgname()` (instance) and `gdk_get_program_class()` (class), and GDK derives that class in `gdk_pre_parse()` from the program name with its first letter upper-cased. Measured on a real window:

```
WM_CLASS(STRING) = "io.github.thezupzup.linthra", "Io.github.thezupzup.linthra"
```

`Io.github.thezupzup.linthra` is not a string any launcher indexes. GNOME Shell happens to find us anyway through its lower-casing fallback (`shell_app_system_lookup_desktop_wmclass`), but that is the last of four heuristics and nothing else is obliged to do it. There was no `StartupWMClass=` for the exact-match path that runs first.

**No window icon at all.** Nothing ever called `gtk_window_set_default_icon_name()`, so GTK attached no `_NET_WM_ICON`:

```
_NET_WM_ICON:  not found.
```

Every task switcher, panel and window list that reads the window's own icon rather than resolving a desktop entry had nothing but a generic placeholder to show.

**And the icon file itself could not be decoded.** This one surprised me and is probably the biggest single cause of the reported symptom. `tool/branding/linthra_icon.svg` opened with a ~300 byte comment between the XML declaration and `<svg>`. An SVG has no magic number, so content sniffing looks for the literal `<svg` and gives up after 256 bytes. Past that, gdk-pixbuf refuses the file outright:

```
tool/branding/linthra_icon.svg -> Couldn't recognize the image file format
```

`gtk_icon_theme_has_icon()` returned 1 (the lookup found the file) and `gtk_icon_theme_load_icon()` still failed. The file parses, validates and renders fine in a browser, which is why nothing caught it. Anything resolving the icon through gdk-pixbuf, GTK itself and GNOME Shell included, fell back to a generic icon. Qt-based shells decode SVG differently and were unaffected, which matches an issue that reproduces on GNOME more readily than on KDE.

## The fix

`linux/runner/my_application.cc`, all four derived from `APPLICATION_ID` so there is no second copy of the id:

| Call | Where | Why there |
| --- | --- | --- |
| `g_set_prgname(APPLICATION_ID)` | `my_application_new()` (unchanged) | before `gtk_init()`, which is where GDK reads the program name |
| `gdk_set_program_class(APPLICATION_ID)` | `my_application_startup()`, after the chain-up | `gtk_init()` resets the program class unconditionally, so earlier is thrown away |
| `gtk_window_set_default_icon_name(APPLICATION_ID)` | same | needs GTK initialised, and must precede window construction |
| `g_set_application_name(AppInfo.name)` | same | otherwise `g_get_application_name()` falls back to the reverse-DNS id |

The ordering is the fix, not a style choice: both GDK calls are undone by `gtk_init()` if they run before the chain-up to `GtkApplication::startup`.

Plus `StartupWMClass=io.github.thezupzup.linthra` in the desktop entry, declaring the pair the runner now stamps, and the icon's leading comment moved inside the root element. The artwork is byte-identical, only the comment moved.

Same window, after:

```
WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
_NET_WM_ICON(CARDINAL) = Icon (48 x 48): ...
```

## Validation

Reproduced end to end with a minimal GtkApplication compiled against GTK 3.24.41, once without the fix and once with it, under `xvfb-run`, with the icon installed at `hicolor/scalable/apps/io.github.thezupzup.linthra.svg` on `XDG_DATA_DIRS`:

* before: WM_CLASS class half wrong, `_NET_WM_ICON: not found`
* after: both WM_CLASS halves are the app id, `_NET_WM_ICON` present, on the first launch and again after close and reopen

Also ran: `check_linux_runner.py` (clean), its 81 tooling tests (green), `ruff check` and `ruff format --check` over `scripts tool tools`, `desktop-file-validate` (no output), and both `-Wall -Werror` under gcc and clang on the four new calls.

`flutter analyze` / `flutter test` were not run here: no Flutter SDK in this environment, and the only Dart file touched is `test/tooling/flatpak_launch_smoke_test.dart`, whose string expectations I cross-checked against the shell script programmatically. CI covers the rest.

Not verifiable from CI: the GNOME Wayland dock and the KDE Plasma Wayland task manager both need a real session. Worth a look on both before this leaves draft.

## Guardrails, so it cannot drift back

`scripts/check_linux_runner.py` now also checks:

* all four identity calls are present, exactly once each
* each takes the `APPLICATION_ID` macro (or `kApplicationName`), not a pasted literal
* each sits in the function where it takes effect
* the two order-dependent ones follow the startup chain-up
* `StartupWMClass=` equals the app id
* the icon's `<svg>` root starts within the first 256 bytes

`scripts/flatpak_launch_smoke.sh` now launches the packaged Flatpak twice and reads WM_CLASS and `_NET_WM_ICON` back off the real window each time, so the identity of the actual shipped artifact is asserted, and close/reopen is covered. Under Xvfb this exercises the sandbox's `--socket=fallback-x11` path.

22 new test cases across `check_linux_runner_test.py` and `flatpak_launch_smoke_test.dart`, each taking a good fixture and breaking exactly one thing.

## Out of scope, untouched

No second app id, no duplicate launcher, no new artwork, no Linux-only rebrand. Android/F-Droid/Play identity, Flatpak permissions (finish-args are byte-identical and still guarded by the existing allow-list), navigation, album grids, Now Playing, MPRIS: all untouched. Nothing here relates to #555 or #556.

https://claude.ai/code/session_01VtodLhcsSefY5c5bx4797V

---
_Generated by [Claude Code](https://claude.ai/code/session_01VtodLhcsSefY5c5bx4797V)_

### DIFF
--- a/.github/workflows/flatpak-build.yml
+++ b/.github/workflows/flatpak-build.yml
@@ -63,15 +63,25 @@ jobs:
       # happens to suggest. The cost is that every tool has to be named here,
       # including ones a dependency would otherwise have pulled in.
       #
-      # elfutils is exactly that case. flatpak-builder only *recommends* it, so
-      # --no-install-recommends leaves the runner without `eu-strip` and
-      # `eu-elfcompress`, and flatpak-builder needs `eu-strip` to split debug
-      # symbols out of every module it builds. Without it the build dies on the
-      # first native module with
+      # Two of these are here because of that, and both failed the same way:
+      # silently absent, then a build error nowhere near the real cause.
+      #
+      # elfutils: flatpak-builder only *recommends* it, so the runner had no
+      # `eu-strip`, which flatpak-builder needs to split debug symbols out of
+      # every module it builds. The build died on the first native module with
       #   Error: module libplacebo: Failed to execute child process "eu-strip"
-      # long before the Linthra module is reached. Installing the package is the
-      # fix rather than passing --disable-debuginfo, which would silently change
-      # what the exported repository contains.
+      # long before the Linthra module was reached. Installing the package is
+      # the fix rather than passing --disable-debuginfo, which would silently
+      # change what the exported repository contains.
+      #
+      # librsvg2-common: nothing in `appstream`/`appstream-compose` depends on
+      # or recommends it, and it is what gives gdk-pixbuf its SVG loader.
+      # Without it, the `appstreamcli compose` that flatpak-builder runs at
+      # finish time cannot read Linthra's scalable icon, drops the component,
+      # and fails the whole build with
+      #   io.github.thezupzup.linthra  E: file-read-error
+      #   general                      E: filters-but-no-output
+      # whose hint report names the icon and "Unrecognized image file format".
       - name: Install host validation and smoke tools
         run: |
           set -euo pipefail
@@ -83,6 +93,7 @@ jobs:
             elfutils \
             flatpak \
             flatpak-builder \
+            librsvg2-common \
             x11-utils \
             xvfb
 
--- a/.github/workflows/flatpak-build.yml
+++ b/.github/workflows/flatpak-build.yml
@@ -58,6 +58,20 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v7
 
+      # --no-install-recommends is deliberate: it keeps the runner close to the
+      # set of tools this job actually needs, rather than whatever a package
+      # happens to suggest. The cost is that every tool has to be named here,
+      # including ones a dependency would otherwise have pulled in.
+      #
+      # elfutils is exactly that case. flatpak-builder only *recommends* it, so
+      # --no-install-recommends leaves the runner without `eu-strip` and
+      # `eu-elfcompress`, and flatpak-builder needs `eu-strip` to split debug
+      # symbols out of every module it builds. Without it the build dies on the
+      # first native module with
+      #   Error: module libplacebo: Failed to execute child process "eu-strip"
+      # long before the Linthra module is reached. Installing the package is the
+      # fix rather than passing --disable-debuginfo, which would silently change
+      # what the exported repository contains.
       - name: Install host validation and smoke tools
         run: |
           set -euo pipefail
@@ -66,6 +80,7 @@ jobs:
             appstream \
             dbus-x11 \
             desktop-file-utils \
+            elfutils \
             flatpak \
             flatpak-builder \
             x11-utils \
--- a/docs/flatpak-ci.md
+++ b/docs/flatpak-ci.md
@@ -39,6 +39,13 @@ Install the host tooling first. On Fedora Atomic, the contributor setup in
 [`flatpak-development.md`](./flatpak-development.md) remains the recommended
 path. On a distribution with native `flatpak-builder`, the core CI sequence is:
 
+> `flatpak-builder` needs **elfutils** (`eu-strip`) to split debug symbols out
+> of each module it builds. Most distributions pull it in with
+> `flatpak-builder`; Debian and Ubuntu only *recommend* it, so a minimal install
+> leaves it out and the build fails on the first native module with
+> `Failed to execute child process "eu-strip"`. The CI job names it explicitly
+> for that reason.
+
 ```bash
 python3 scripts/check_linux_runner.py
 python3 test/tooling/check_linux_runner_test.py
@@ -74,8 +81,8 @@ flatpak build-update-repo repo-ci
 bash ../scripts/flatpak_launch_smoke.sh repo-ci
 ```
 
-The launch command requires `xvfb-run`, `xwininfo` and `dbus-run-session`, the
-same tools installed by the CI job. It intentionally refuses to run if Linthra
+The launch command requires `xvfb-run`, `xwininfo`, `xprop` and
+`dbus-run-session`, the same tools installed by the CI job. It intentionally refuses to run if Linthra
 is already installed for the current user or system-wide, so a local smoke can
 never launch or remove a contributor's existing installation. Use a clean test
 user/environment for this final step when Linthra is already installed.
--- a/docs/flatpak-ci.md
+++ b/docs/flatpak-ci.md
@@ -39,12 +39,19 @@ Install the host tooling first. On Fedora Atomic, the contributor setup in
 [`flatpak-development.md`](./flatpak-development.md) remains the recommended
 path. On a distribution with native `flatpak-builder`, the core CI sequence is:
 
-> `flatpak-builder` needs **elfutils** (`eu-strip`) to split debug symbols out
-> of each module it builds. Most distributions pull it in with
-> `flatpak-builder`; Debian and Ubuntu only *recommend* it, so a minimal install
-> leaves it out and the build fails on the first native module with
-> `Failed to execute child process "eu-strip"`. The CI job names it explicitly
-> for that reason.
+> Two host packages are easy to miss because nothing hard-depends on them, and
+> both fail late and confusingly:
+>
+> * **elfutils** (`eu-strip`), which `flatpak-builder` uses to split debug
+>   symbols out of each module. Debian and Ubuntu only *recommend* it, so a
+>   minimal install leaves it out and the build fails on the first native module
+>   with `Failed to execute child process "eu-strip"`.
+> * **librsvg2-common**, the gdk-pixbuf SVG loader. Without it the
+>   `appstreamcli compose` step at finish time cannot read Linthra's scalable
+>   icon and fails the build with `file-read-error` plus
+>   `filters-but-no-output`. Nothing in `appstream` depends on or recommends it.
+>
+> The CI job names both explicitly for that reason.
 
 ```bash
 python3 scripts/check_linux_runner.py
--- a/docs/linux-desktop.md
+++ b/docs/linux-desktop.md
@@ -286,9 +286,14 @@ widget.
 
 * `scripts/check_linux_runner.py` — the committed runner still matches the app's
   identity (`APPLICATION_ID` = Android's `applicationId`, `BINARY_NAME` =
-  the package name, window title = `AppInfo.name`), the window metrics are
-  sane, the offline SQLite seam is wired, and nothing under `linux/` hardcodes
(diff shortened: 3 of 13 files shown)


TheZupZup:
CI status on `db129cf`: everything green except **Build and launch Flatpak**, and that one is not this PR's.

Green, including the two that actually exercise this change:

* **Build Linux desktop**: `flutter build linux --release` compiles the runner with the new `gdk_set_program_class()` / `gtk_window_set_default_icon_name()` / `g_set_application_name()` calls under the project's real `-Wall -Werror`, and the job also ran `check_linux_runner.py`, its 81 tooling tests, and `desktop-file-validate` over the entry with the new `StartupWMClass=`.
* **Flutter checks**: format, analyze and `flutter test`, including the updated `flatpak_launch_smoke_test.dart`.
* Ruff, secret and privacy scan, repository integrity guard tests, debug APK, both sponsor-state validations.

### The Flatpak failure

```
compressing debuginfo in: .../flatpak-builder-ci/files/lib64/libplacebo.so.360
Warning: eu-elfcompress not installed, will not compress debuginfo
stripping .../libplacebo.so.360 to .../lib/debug/lib64/libplacebo.so.360.debug
Error: module libplacebo: Failed to execute child process "eu-strip" (No such file or directory)
```

It dies building the `libplacebo` module, before the app module, so it never reaches anything this PR touches: not the runner, not the desktop entry, not the icon, not the launch smoke.

**It is red on `main` too, identically.** [Run 11](https://github.com/TheZupZup/Linthra/actions/runs/33777902095) on `main` at `e63d1c7` (the merge of #549) fails on the same module, same step, same missing binary. Every completed run of this workflow so far is a failure or a cancellation, so it has never been green.

### Root cause and proposed patch

`eu-strip` and `eu-elfcompress` come from `elfutils`, which `flatpak-builder` only *recommends*. The host tool install step uses `--no-install-recommends`, so `elfutils` never lands on the runner and flatpak-builder cannot strip debuginfo out of any module it builds:

```diff
--- a/.github/workflows/flatpak-build.yml
+++ b/.github/workflows/flatpak-build.yml
@@ -66,6 +66,7 @@
             appstream \
             dbus-x11 \
             desktop-file-utils \
+            elfutils \
             flatpak \
             flatpak-builder \
             x11-utils \
```

The alternative, `--disable-debuginfo` on the `flatpak-builder` calls, works too but changes what the exported repo contains, so adding the package is the smaller and more faithful fix.

I have not applied it here. It is a pre-existing CI gap in code this PR does not touch, and I would rather not widen an app-identity PR into a workflow fix without you deciding. Happy to push it into this branch or open it as its own PR, whichever you prefer.

No re-run: the failure is a deterministic missing binary and the base branch already shows the same thing, so a re-run would only burn eight minutes of runner time. Once the `elfutils` fix lands somewhere, this job should get past libplacebo and reach the new `WM_CLASS` / `_NET_WM_ICON` assertions on the packaged app, which are the part of this PR that CI cannot currently prove.

---

TheZupZup:
Following up on my earlier comment: the `elfutils` fix landed here, and two more blockers surfaced behind it. **Build and launch Flatpak is now green**, and the workflow has never been green before, so this is the first time it has actually built and launched the package.

Three pre-existing blockers, each diagnosed from the real logs, in the order they appeared:

**1. `eu-strip` missing** (`df15e35`). `flatpak-builder` only *recommends* `elfutils`, and the job installs with `--no-install-recommends`. It died on the first native module.

**2. libplacebo installed to `/app/lib64`** (`78fa7e1`). With the build reaching further, mpv failed its configure with `Dependency "libplacebo" not found`. libplacebo had built fine but installed to the wrong prefix: flatpak-builder passes meson only `--prefix=/app` (the module's "User defined options" in the log list `prefix` and nothing else), so meson used its own default libdir, which is `lib64` in this SDK because the freedesktop base carries a real `/usr/lib64`. That is not on flatpak-builder's `PKG_CONFIG_PATH`. The autotools modules are unaffected and libass installs its `.pc` to `/app/lib/pkgconfig` in the same run, which is what made the cause obvious. Pinned `-Dlibdir=lib` on both meson modules, not just the one that failed: the bundled libmpv has to be on the sandbox library path for the dlopen `media_kit` does at startup.

**3. No SVG loader for `appstreamcli compose`** (`efb6b73`). The build then completed, including the Flutter module, and died at finish time with `file-read-error` plus `filters-but-no-output`. Reproduced locally on the same appstreamcli 1.0.2; the hint report names it exactly:

```
fname: //share/icons/hicolor/scalable/apps/io.github.thezupzup.linthra.svg
msg: Unrecognized image file format
```

`appstreamcli compose` rasterises the icon through gdk-pixbuf, and the SVG loader is in `librsvg2-common`, which nothing in the appstream stack depends on or recommends. Confirmed causally by removing only that loader from a working setup: identical errors, identical hint.

Worth recording, because it corroborates this PR's main diagnosis: the same experiment shows the icon change in this branch was independently required. The pre-#554 icon, whose `<svg>` root sat past the 256-byte sniffing window, fails this same compose step with the identical `file-read-error` even when the loader is present. Both fixes are needed.

No `--disable-debuginfo`, no manifest source or version changes, no sandbox permission changes. The smoke logged the installed permission set as `ipc network fallback-x11 pulseaudio wayland x11 dri`, unchanged.

### What the green run actually did

```
Running appstreamcli compose ... Success!
Finishing app
Exporting share/applications/io.github.thezupzup.linthra.desktop
Exporting share/icons/hicolor/scalable/apps/io.github.thezupzup.linthra.svg
Exporting share/metainfo/io.github.thezupzup.linthra.metainfo.xml
Exporting io.github.thezupzup.linthra to repo
Flatpak build and repository export completed successfully.
Installed io.github.thezupzup.linthra from local repository

PASS: packaged io.github.thezupzup.linthra opened a Linthra window (first launch).
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
PASS: Linthra window carries _NET_WM_ICON (_NET_WM_ICON(CARDINAL) = Icon (48 x 48):).
PASS: packaged io.github.thezupzup.linthra opened a Linthra window (reopen after close).
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
PASS: Linthra window carries _NET_WM_ICON (_NET_WM_ICON(CARDINAL) = Icon (48 x 48):).
```

So the identity assertions this PR adds are no longer theoretical: the real packaged Flatpak was installed, launched, and both halves of its `WM_CLASS` are the application id, with the Linthra icon attached to the window, on a fresh launch and again after close and reopen.

All 16 checks on `efb6b73` are green.

---

PR #541, opened 2026-09-01
feat(linux): add AppStream metainfo for software centres #435

AzazelSensei:
Adds the AppStream listing from #435.

`linux/packaging/io.github.thezupzup.linthra.metainfo.xml` uses the same application id as the desktop entry and icon. Both Flatpak manifests install it to `/app/share/metainfo/`. The runner checker holds the component id, launchable desktop-id, and stock icon to that identity.

The description only covers local files and Jellyfin / Navidrome / Subsonic / Plex streaming. I left screenshots out until there are Linux captures.

`python3 test/tooling/check_linux_runner_test.py` is green. `appstreamcli validate` is wired into `scripts/verify_linux.sh` and the Linux desktop CI job.

Closes #435

PR #563, opened 2026-09-06
release: prepare Linthra v0.2.5

TheZupZup:
Release preparation for v0.2.5. No new product features, no refactors, no architecture or permission changes.

Cut from `main` at `88a61eac9cd538bf4ff2d6ee98278ec679eda394`.

## Release summary

v0.2.5 is the desktop release. Everything that landed since v0.2.4 (#537, #540, #541, #542, #543, #544, #550, #557, #558, #559, #560, #561, #562) is either Linux desktop work or Android user-visible fixes.

## Main user-visible changes

- Persistent navigation shell on wide Linux windows, with per-destination stacks and correct keyboard traversal
- Adaptive layouts driven by the width a view is actually given, no-ops below 600 px
- Denser album grid (5 columns in a 1280-wide window, 7 at 1920, 9 at 2560) and a 3/4/5-column artist grid at the same widths
- Album and artist detail get a persistent side pane at expanded widths
- Now Playing puts the cover beside the metadata, and lyrics open next to it instead of replacing it
- Songs, playlists, downloads and settings cap and centre their column
- Navigation state survives resizes across breakpoints
- Transport controls on the now-playing bar, shedding controls as the window narrows
- Folders promoted to a top-level destination
- Linux WM_CLASS matches the app id exactly, from the runner, so it holds for both Linux builds. The desktop entry, raster icons and `_NET_WM_ICON` are Flatpak-only: the native x64 archive ships no desktop integration assets
- Music folder picking through `xdg-desktop-portal`, credentials through the Secret portal (no finish-arg needed), configured-server networking in the Flatpak
- Android: device-wide music access is an explicit, transactional choice
- Android Auto serves Songs in bounded pages, so large libraries no longer spin forever
- Jellyfin availability is tracked separately from configuration

## Version / versionCode

| Field | Value |
| --- | --- |
| Tag (not created yet) | `v0.2.5` |
| `pubspec.yaml` | `0.2.5+205999` |
| `AppInfo._devVersionName` | `0.2.5` |
| Fastlane changelog | `fastlane/metadata/android/en-US/changelogs/205999.txt` |
| F-Droid `CurrentVersion` / `CurrentVersionCode` | `0.2.5` / `2059993` |
| Per-ABI | `2059991` / `2059992` / `2059993` |

`205999` comes from `scripts/prepare_release_bump.py`, which mirrors `tool/version_from_tag.dart`. Nothing was hand-computed and no historical versioning rule was touched.

`CurrentVersion` in the F-Droid metadata was stale at `0.1.10` — the last four release bumps left that file alone. The bump script updates it, and both F-Droid guardrail tests are green on the new values.

## F-Droid license cutover

Advancing `CurrentVersion` to 0.2.5 makes that metadata point at the first release cut from an AGPL source commit, and `docs/license-transition.md` requires the `License:` field to move to `AGPL-3.0-or-later` **in the same change** that does so. This PR does that.

`test/tooling/license_metadata_guardrails_test.dart` moves to the post-cutover state its own doc comment anticipated: it asserts the current license and the absence of the old one, plus a new case holding the historical `Builds` entries to their pre-transition versions so a future `CurrentVersion` bump can never relabel them. `license-transition.md` and `fdroid-submission.md` §9 are updated to record the cutover as done and to name the external `fdroiddata` recipe and submission text as what is still outstanding.

No old build commit, tag, release note or artifact is relabelled.

## What's New

Yes, updated. `WhatsNewSection.releaseNotes` carries six user-facing bullets: the desktop shell, adaptive layouts and grids, the now-playing transport and Folders, resize behaviour and Linux app identity, explicit device-wide music access plus the Android Auto fix, and honest Jellyfin availability. No internal jargon, no CI plumbing, no unshipped work. The version label beside it reads from `AppInfo`, so it shows `Version 0.2.5` automatically.

Four surfaces, each written for its own audience:

- in-app What's New: short and user-facing
- `fastlane/metadata/android/en-US/changelogs/205999.txt`: four lines, Android-first, 379 characters (F-Droid truncates `whatsNew` at 500)
- `docs/release-notes/v0.2.5.md`: the longer GitHub Release body
- a `0.2.5` entry in the AppStream `<releases>` list

## Validation

Local, against the pinned Flutter 3.44.7 from `scripts/setup_flutter.sh`:

| Check | Result |
| --- | --- |
| `flutter pub get --enforce-lockfile` | pass |
| `dart format --output=none --set-exit-if-changed .` | pass, 932 files, 0 changed |
| `flutter analyze` | pass, no issues |
| `flutter test` | pass, 4049 tests |
| `python3 -m pytest test/tooling/` | pass, 437 tests + 270 subtests |
| `ruff check .` / `ruff format --check .` | pass |
| `./scripts/release_preflight.sh v0.2.5` | `OK: v0.2.5 matches pubspec.yaml version 0.2.5+205999` |
| `python3 scripts/check_linux_runner.py` | `Linux runner configuration OK.` |
| `./scripts/check_secrets.sh` | pass |
| `xmllint --noout` on the metainfo | well-formed |

CI on the head commit covers the three things this environment could not build (no Android SDK, and `dl.google.com` / `sqlite.org` / the Flathub runtimes are all blocked by the egress policy). All green: Flutter checks, Build debug APK, Build Linux desktop, Build and launch Flatpak, Repository integrity, Security surface guard, Secret & privacy scan, Sponsor simulation.

The packaged Flatpak smoke test confirms the window identity end to end on both a first launch and a reopen:

```
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
PASS: Linthra window carries _NET_WM_ICON (_NET_WM_ICON(CARDINAL) = Icon (48 x 48):)
```

That job failed once on the first commit of this branch and passed on the third; see the comment thread below. Nothing in this diff can affect an X window property, and the check has no history against #559 to #562 because `flatpak-build.yml` is path-filtered, so it looks intermittent rather than regressed. Worth a look on its own, not a v0.2.5 blocker.

## Known limitations

`scripts/check_release_bump_files.sh` allows only pubspec, `app_info.dart`, the Fastlane changelog, the F-Droid metadata and `docs/release-notes/**`. This PR also touches `whats_new_section.dart`, the metainfo, the two license documents and the license guardrail test, which that guard would reject. The CI job is scoped to `release/*` branches, so it is skipped here and CI stays green. If you would rather keep the strict version-only shape, the What's New change can move to its own PR the way #533 did for v0.2.4 — though the license cutover genuinely has to ride with the version bump.

`README.md` still says the current stable is v0.2.4, matching repo practice of bumping that badge after the release is actually published.

## Scope

No new features. No provider, playback or navigation architecture changes. No Flatpak or Android permission changes. No dependency, Gradle or AGP changes. No test was weakened or removed — the license guardrail was advanced to its next state and gained a case.

No tag was created, no GitHub Release was drafted, and this PR is not merged.

https://claude.ai/code/session_01QjWsFYYxLtBz8uNkzUrEy5


TheZupZup:
**Build and launch Flatpak** is red on `d7438a9`, and I don't think it belongs to this PR. Writing down what I found so it isn't lost.

The Flatpak itself builds and exports cleanly, icons included:

```
Exporting share/icons/hicolor/scalable/apps/io.github.thezupzup.linthra.svg
Exporting share/icons/hicolor/48x48/apps/io.github.thezupzup.linthra.png
Exporting share/icons/hicolor/64x64/apps/io.github.thezupzup.linthra.png
Exporting share/icons/hicolor/256x256/apps/io.github.thezupzup.linthra.png
Exporting share/icons/hicolor/128x128/apps/io.github.thezupzup.linthra.png
Flatpak build and repository export completed successfully.
```

It fails in `scripts/flatpak_launch_smoke.sh`, on the first launch:

```
PASS: packaged io.github.thezupzup.linthra opened a Linthra window (first launch).
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
FAIL: the Linthra window carries no _NET_WM_ICON (first launch)
```

Why it isn't this PR's: the diff here is the version bump, the Fastlane changelog, the F-Droid metadata, `docs/release-notes/v0.2.5.md`, a const list of strings in `whats_new_section.dart`, and one `<release>` line in the metainfo. None of that can affect an X window property. The two files that reach the Flatpak are `pubspec.yaml` (version only) and the metainfo, and the metainfo composed fine ("Running appstreamcli compose ... Success!").

What makes this awkward to attribute is the path filter. `flatpak-build.yml` only runs on `flatpak/**`, `linux/**`, `pubspec.yaml` and friends, so its last run on `main` was #558 at `3c885d0`, which was green. #559, #560, #561 and #562 all landed after that without touching a trigger path, so the Flatpak smoke test has not run against any of them. This PR is the first thing to re-trigger it, twelve hours and four commits later. So the failure is either a regression introduced by one of those four, or a race in the smoke test that #558's two green runs got past.

I can't tell which from here. I have no permission to re-run the job or dispatch the workflow on `main` (both come back 403), and the Flatpak toolchain isn't reachable in my environment, so I can't reproduce it locally either. There's no existing fix to port.

One thing I noticed while reading the script, offered as a lead rather than a diagnosis: `window_line()` is `xwininfo -root -tree | grep -F '"Linthra":' | head -n 1`, so it takes whichever matching window appears first in the tree dump. Under bare Xvfb with no window manager, a second window carrying the same title would be picked up the same way, and a GTK popup inherits WM_CLASS but does not get the default icon list. #559 and #562 did change what the shell puts on screen. That would explain WM_CLASS passing while `_NET_WM_ICON` is absent, and it would also explain why the same check was green twice in a row before those landed. Worth checking before assuming the icon fix regressed.

I'm deliberately not touching the smoke script here. If it is a real regression, weakening the check would hide it, and if it's a race, the fix belongs in its own PR rather than riding along on a version bump.

One thing this does bear on for the release: the notes and the in-app What's New both say the Linux window shows the right name and icon. The name half is confirmed green (WM_CLASS passes). If the icon half turns out to be a genuine regression, that wording should be revisited before tagging. Flagging it rather than quietly editing it, since we don't know yet which it is.

I'll keep watching the PR.

---

TheZupZup:
Closing this out: **Build and launch Flatpak passed** on `c14f947`, cleanly, on both launches.

```
PASS: packaged io.github.thezupzup.linthra opened a Linthra window (first launch).
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
PASS: Linthra window carries _NET_WM_ICON (_NET_WM_ICON(CARDINAL) = 	Icon (48 x 48):).
PASS: packaged io.github.thezupzup.linthra opened a Linthra window (reopen after close).
PASS: WM_CLASS(STRING) = "io.github.thezupzup.linthra", "io.github.thezupzup.linthra"
PASS: Linthra window carries _NET_WM_ICON (_NET_WM_ICON(CARDINAL) = 	Icon (48 x 48):).
```

So the `_NET_WM_ICON` failure on `d7438a9` was intermittent, not a regression from #559 to #562. Same branch, same smoke script, and nothing between those commits touches the runner, the manifest or the icons.

Two caveats worth keeping:

The middle runs never finished. `flatpak-build.yml` sets `cancel-in-progress: true` and the job takes about ten minutes, so my pushes for the review fixes cancelled runs 21 and 22 before they completed. The evidence is one fail and one pass, not one fail and three passes, which is thin for calling a rate.

And the underlying hazard is still there. `window_line()` is `xwininfo -root -tree | grep -F '"Linthra":' | head -n 1`, so it takes whichever matching window the tree dump lists first, and it reads `_NET_WM_ICON` immediately after finding one. In the failing run only ~3 ms separated the WM_CLASS read from the icon read. If a second window titled Linthra can exist even briefly, or if the icon property lands just after the window is mapped, the check can lose that race. Worth hardening independently, with a short retry on the property or by pinning the search to a real toplevel. I'm not touching it in a version bump.

Everything on the PR is green and `mergeable_state` is clean.

---

Issue #576, opened 2026-09-08
[P1][Security] Design least-privilege remote media access for external playback


TheZupZup:
## Priority
P1 — defense-in-depth and architectural remediation. Do not block the emergency containment release on this work.

## Goal
Reduce the authority and lifetime of credentials or access capabilities handed to external playback devices. A receiver should receive only the access necessary to play the selected media, not broader account access when a narrower provider-supported mechanism is available.

## Scope
- Review the external-playback media resolution contracts for supported remote providers.
- Document which providers support short-lived, resource-scoped capabilities, signed URLs, or another suitable delegation mechanism.
- Evaluate an application-managed or server-supported media proxy only after reviewing its authentication, authorization, network exposure, lifecycle, and operational costs.
- Avoid inventing client-side token restrictions that the media server does not actually enforce.
- Preserve compatibility with self-hosted deployments and document provider limitations rather than silently weakening the security model.
- Ensure secrets are not unnecessarily exposed through logs, diagnostics, errors, metadata, or persistent storage.
- Define appropriate expiry, revocation, and cleanup behavior for any newly introduced capabilities.
- Keep this work separate from the receiver-authentication requirement; a scoped URL is not a substitute for authenticating the destination.
- Coordinate sensitive findings and implementation details through the private security advisory.

## Likely code areas
- `lib/core/services/cast/`
- `lib/core/sources/jellyfin/`
- `lib/core/sources/subsonic/`
- Media resolution interfaces and provider capabilities
- Credential handling, tests, and security documentation

## Acceptance criteria
- [ ] Provider capabilities and limitations are documented.
- [ ] The selected design minimizes access scope and lifetime where supported.
- [ ] No broader credentials are introduced into external-playback payloads as a fallback.
- [ ] Existing playback, server compatibility, and lifecycle behavior are covered by tests.
- [ ] Any proxy design receives a separate security and operational review before implementation.
- [ ] The final design is reviewed privately before relevant security details are disclosed.

## Coordination
Related to #575. This is an additional security layer, not a reason to delay #572–#574 or to restore Cast before receiver authentication is approved. The underlying security report remains private; see the resolved contact request #545 for coordination history.

Issue #575, opened 2026-09-08
[P1][Security] Restore Cast only with verified receiver trust


TheZupZup:
## Priority
P1 — priority security remediation, immediately after emergency containment. This must not delay the maintenance release.

## Goal
Restore Cast support only after the receiver's identity has been cryptographically authenticated and the complete media handoff path has passed an independent security review. Preserve Linthra's open-source and F-Droid-compatible architecture where practical, without weakening security requirements to maintain compatibility.

## Scope
- Evaluate a maintained implementation, an auditable fork, or a narrowly scoped replacement rather than assuming an unreviewed dependency update is sufficient.
- Establish an explicit receiver-authentication contract and a fail-closed readiness boundary.
- Require successful authentication before any media handoff or other operation that requires a trusted receiver.
- Bind the authenticated connection to the receiver selected by the user and handle connection changes, cancellation, and session lifecycle safely.
- Review the implementation's trust model, protocol compatibility, dependency provenance, and supported platforms with the reporter in the private advisory.
- Add positive interoperability tests with supported devices and negative authentication tests using controlled, non-production fixtures.
- Review the credential exposure and authorization properties of the resulting media handoff before restoring production availability.
- Retain the temporary production safeguard until the replacement and its tests are reviewed and approved. No runtime bypass or insecure fallback is acceptable.

## Likely code areas
- `lib/core/services/cast/`
- `lib/features/player/cast/`
- Production application wiring
- Dependency manifests and lockfiles
- Cast protocol integration tests and security documentation

## Acceptance criteria
- [ ] The receiver trust model and implementation choice are documented and privately reviewed.
- [ ] Authentication failures, identity mismatches, and incomplete authentication cannot reach trusted readiness.
- [ ] No credential-bearing media handoff can occur before the required trust checks succeed.
- [ ] Controlled negative tests and supported-device interoperability tests pass.
- [ ] Dependencies and packaged artifacts are reviewed for the approved implementation.
- [ ] Production restoration is a separate, reviewed change and does not reintroduce an insecure fallback.
- [ ] The reporter is offered an opportunity to review the remediation through the private advisory.

## Disclosure coordination
The complete technical requirements, protocol-specific evidence, and security test scenarios belong in the private advisory. Do not reproduce them in public comments or PR descriptions before coordinated disclosure.

Related: #572, #573, #574, and the resolved private-contact request #545.

Issue #574, opened 2026-09-08
[P0][Security] Ship and verify an emergency maintenance release


TheZupZup:
## Priority
P0 — ship the containment as soon as it passes security review and release validation. Do not wait for the permanent Cast replacement or unrelated roadmap work.

## Goal
Deliver the temporary security safeguard to existing users through Linthra's supported distribution channels and confirm that the released artifacts contain it.

## Requirements
- Coordinate the release decision, affected-version assessment, and disclosure timing in the private security advisory.
- Select the next appropriate version from the current release state; do not assume a version number before checking tags, branches, and distribution metadata.
- Merge the reviewed containment and regression coverage from #572 and #573.
- Run the required analysis, tests, packaging checks, and real-device smoke tests for the affected production platforms.
- Build the release artifacts from the reviewed commit and verify that the production Cast safeguard is present in the actual distributed binaries.
- Record the release commit, version, artifact names, and SHA-256 digests for auditability.
- Publish through the applicable GitHub, F-Droid, and other supported release channels, following each channel's normal signing and update requirements.
- Verify that the distribution automation succeeded and that users can obtain the new version; do not treat a successful local build or tag creation as proof of publication.
- Coordinate appropriate user guidance, disclosure, and any credential-rotation recommendations with the private advisory. Do not claim that exploitation or credential theft occurred without evidence.
- Keep unrelated feature changes out of this maintenance release.

## Likely code areas
- Release/version metadata
- `.github/workflows/`
- F-Droid and other distribution packaging metadata
- Release notes and security coordination
- Android/iOS production build configuration

## Acceptance criteria
- [ ] Reviewed containment and required regression tests are merged.
- [ ] Release CI and relevant production smoke tests pass.
- [ ] Final artifacts are tied to a reviewed commit and recorded hashes.
- [ ] Applicable distribution channels receive the maintenance release.
- [ ] Installation/update verification is completed where practical.
- [ ] Public messaging is coordinated with the private advisory.
- [ ] The long-term remediation remains tracked independently; this release does not claim to restore Cast.

## Coordination
Blocked on #572 and #573. Private security details must remain in the advisory associated with #545 until coordinated disclosure.

PR #586, opened 2026-09-08
Cast security follow-ups: verify the shipped containment, receiver trust, and what a handoff delegates (#574, #575, #576)

TheZupZup:
Three issues, one commit each, plus a follow-up commit for the review findings. Casting stays off in shipped builds: nothing here restores it, and nothing new is wired into production.

## #574 — verify and record the maintenance release

The containment shipped in v0.2.6, but everything checking it looked at the source tree. For a security release the question is whether the safeguard is in the file people install.

- `scripts/verify_release_containment.py` reads the compiled Dart inside an APK, an AAB or the Linux tarball and checks for the containment message and `UnavailableCastService`, plus the absence of the live-cast strings tree shaking drops while nothing constructs them. Prints each artifact's SHA-256, fails closed, and says plainly in its docstring that it is a byte search rather than a proof.
- It runs before an artifact is uploaded anywhere (the Android release build and the Linux packaging job), and again on the published assets during a stable publication, with the digests recorded in the job summary. Its own unit tests run in CI on synthetic artifacts.
- `docs/release-artifact-verification.md` records the v0.2.6 run: commit `f1d61f9`, version `0.2.6+206999`, and every published asset with its SHA-256, plus where each distribution channel stands.

I ran it against the six published v0.2.6 assets while writing this: all of them carry the containment, none carries a live-cast marker. Those are the digests in the doc.

## #575 — receiver trust contract and fail-closed gate

The app's half of the restoration, built ahead of the implementation choice so reviewing the eventual transport is about the protocol rather than our plumbing.

- `CastReceiverAuthenticator` is handed the session to check and returns an identity naming both the device and that connection, or throws. The connection half matters: Cast's device authentication is a challenge answered on a connection, so a proof gathered elsewhere describes a different conversation than the one the media would go to. An inconclusive check is a failure. The shipped `UnverifiedCastReceiverAuthenticator` refuses every receiver.
- `TrustGatedCastTransport` hands out a session only after the receiver authenticated and the identity matches both the device the user picked and the session it will use. Any failure closes the session (bounded, so an unresponsive receiver cannot hold the refusal back), the media handoff refuses again on its own, and trust ends with the connection it was proved over — a close from here or the receiver dropping.
- `DefaultCastService` surfaces a refusal's own secret-free message instead of a generic "couldn't connect", so a security refusal does not read as a flaky network.
- Containment still comes first: the gate authenticates over a connection the delegate opened, so a contained transport refuses before trust is even asked about. There is a test for that, so this cannot become a second route to a receiver.
- Negative tests cover a refusal, a device authenticating as another device, a proof made over another connection, a handshake that never finishes, a session dropping or dying mid-handshake, a drop after a successful check, an authenticator throwing something unexpected, and a receiver that will not close cleanly or at all. No refusal message carries anything from the handshake.
- `docs/cast-receiver-trust.md` has the trust model, implementation options and trade-offs, the negative/device test matrix, and the restoration checklist.

## #576 — what a cast handoff delegates

- `CastMediaAccess` makes each source declare what casting one of its tracks hands over: a capability scoped to the item, or the account credential, with scope, lifetime and revocation.
- Jellyfin and Subsonic both declare `accountCredential` / `account`, because neither issues a per-item signed URL. No client-side restriction the server does not enforce, and no broader credential as a fallback.
- A source that declares nothing reads as account-level exposure, so an omission overstates rather than understates.
- `CastMediaResolver.accessFor` answers without resolving or touching the network; routing reports the source that would actually cast the track.
- `docs/cast-media-access.md` has the per-server matrix, what each server would need for a scoped capability, and the on-device media proxy option with its real costs, deferred to its own security and operational review.

## Checks

- `flutter analyze` clean, full `flutter test` green (4249 tests), `dart format` clean.
- `ruff check` / `ruff format --check` clean over `scripts tool tools`.
- `scripts/check_cast_containment.py` still passes, and `test/tooling/verify_release_containment_test.py` passes.

Security-sensitive details stay in the private advisory; nothing about the report is in the code, the docs, or these commits.

### DIFF
--- a/.github/workflows/android-release-build.yml
+++ b/.github/workflows/android-release-build.yml
@@ -162,6 +162,18 @@ jobs:
           # manual one.
           ref: ${{ inputs.release_tag }}
 
+      # The release tooling from the revision that is RUNNING, kept beside the
+      # build source rather than replacing it. A run rebuilding an older tag
+      # checks that tag's tree out above, which may predate the containment
+      # verifier; the check still has to happen, so the script comes from here
+      # and the expected message from the tag's own cast_containment.dart.
+      - name: Check out the release tooling from this workflow's revision
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ github.sha }}
+          path: .release-tooling
+          sparse-checkout: scripts
+
       # Decide, in one place, how this run should sign, whether the tag is a
       # pre-release, and how to label/name the artifacts.
       #   trigger:    manual | tag
@@ -494,17 +506,22 @@ jobs:
         run: |
           set -euo pipefail
 
-          # A run rebuilding an OLD release tag checks out that tag's tree,
-          # which may predate this script (and the containment it looks for).
-          # That path is documented and must keep working, so say plainly that
-          # nothing was checked rather than failing the rebuild on a file the
-          # tag never had. Every current release builds a tree that has it.
-          if [ ! -f scripts/verify_release_containment.py ]; then
-            echo "::warning::This tag predates scripts/verify_release_containment.py, so the artifacts were not checked for the Cast containment. Verify the published assets by hand if the tag is expected to carry it (see docs/release-artifact-verification.md)."
+          source_file="lib/core/services/cast/cast_containment.dart"
+
+          # The one case with nothing to check: a tag from before the
+          # containment existed. Then there is no message to look for, and the
+          # documented rebuild path for such a tag must not fail on its absence.
+          # Anything else is checked — with the verifier from this workflow's
+          # revision, against the built tree's own message — because "the tag is
+          # old" must never become a way to replace public assets unverified.
+          if [ ! -f "$source_file" ] || ! grep -q 'static const String userMessage' "$source_file"; then
+            echo "::warning::This tag predates the Cast containment, so its artifacts carry nothing to check. See docs/release-artifact-verification.md."
             exit 0
           fi
 
-          python3 scripts/verify_release_containment.py dist/*.apk dist/*.aab
+          python3 .release-tooling/scripts/verify_release_containment.py \
+            --containment-source "$source_file" \
+            dist/*.apk dist/*.aab
 
       - name: Upload release APK
         uses: actions/upload-artifact@v7
--- a/.github/workflows/android-release-build.yml
+++ b/.github/workflows/android-release-build.yml
@@ -471,6 +471,19 @@ jobs:
 
           echo "Verified: all APK versions match the tag-derived version."
 
+      # Casting is withheld from shipped builds while a reported security issue
+      # is resolved (see lib/core/services/cast/cast_containment.dart). The
+      # source tripwire and the containment tests both run against the tree; this
+      # reads the compiled Dart inside the artifacts this job is about to upload,
+      # so a build that lost containment somewhere between the checkout and the
+      # APK never becomes a Release asset. It also prints each artifact's SHA-256
+      # for the release record. Fails closed — see the script's docstring for
+      # what it can and cannot see.
+      - name: Verify the built artifacts carry the Cast containment
+        run: |
+          set -euo pipefail
+          python3 scripts/verify_release_containment.py dist/*.apk dist/*.aab
+
       - name: Upload release APK
         uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/android-release-build.yml
+++ b/.github/workflows/android-release-build.yml
@@ -150,6 +150,17 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+        with:
+          # A manual run targeting an existing release tag must build *that
+          # tag's* commit. The dispatch's own `--ref` only says which copy of
+          # this workflow file to run (publish-stable-release.yml dispatches
+          # with `--ref main`), so without this the artifacts would come from
+          # whatever `main` is at that moment — which is not necessarily the
+          # reviewed, tagged commit the Release claims to be built from.
+          # Empty for every other trigger, which is checkout's own default: the
+          # pushed tag for a tag build, the dispatched branch for a plain
+          # manual one.
+          ref: ${{ inputs.release_tag }}
 
       # Decide, in one place, how this run should sign, whether the tag is a
       # pre-release, and how to label/name the artifacts.
--- a/.github/workflows/android-release-build.yml
+++ b/.github/workflows/android-release-build.yml
@@ -493,6 +493,17 @@ jobs:
       - name: Verify the built artifacts carry the Cast containment
         run: |
           set -euo pipefail
+
+          # A run rebuilding an OLD release tag checks out that tag's tree,
+          # which may predate this script (and the containment it looks for).
+          # That path is documented and must keep working, so say plainly that
+          # nothing was checked rather than failing the rebuild on a file the
+          # tag never had. Every current release builds a tree that has it.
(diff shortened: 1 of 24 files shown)


TheZupZup:
All six review findings were real and are fixed in `1e5bb89`.

- **Linux tarball verified before it is public.** The packaging job in `linux-desktop-build.yml` now checks out the released tag and verifies the archive before `gh release upload`, so a tarball that failed the check never becomes downloadable from the already-public Release. The Android build already did this.
- **Proof bound to its connection.** `CastReceiverAuthenticator.authenticate` now takes the session, and `CastReceiverIdentity` names the connection the proof was made over; the gate checks device *and* connection. A proof made elsewhere is refused.
- **Readiness drop during the handshake fails the check.** A `false` after a `true` is the session ending; a `false` before is still just "not ready yet", the way `DefaultCastService` reads it.
- **Trust ends with the connection.** The wrapper follows the delegate's lifetime and revokes on drop, error or completion, so a track resolution racing the receiver dropping cannot reach `loadMedia`.
- **Bounded cleanup.** The close on a failed check is time-bounded, so an unresponsive receiver cannot hold the refusal back and strand the UI in "connecting".
- **Refusals keep their message.** `DefaultCastService` catches `CastReceiverTrustException` on both connect and handoff and surfaces its secret-free message, instead of a generic "couldn't connect" that would read as a flaky network.

Tests added for each: a proof over another connection, a drop mid-handshake, a drop after a successful check, a receiver that never closes, and both service-boundary messages. Full suite green (4249 tests), analyze and format clean.


---

TheZupZup:
Second round, four findings, all real and fixed in `d0c9964`.

- **Readiness observed during the handshake is carried over.** It was local to `_authenticate`, so a session that went ready and then dropped in the gap before the wrapper subscribed looked like "not ready yet" to the wrapper, and trust survived a dead connection. The check now reports what it saw and the wrapper starts from that. Test: a silent drop between the check and the handoff refuses.
- **Readiness errors no longer escape.** The wrapper turns an error on the delegate's stream into a `false` and closes, so the service follows its ordinary session-lost teardown instead of getting an unhandled async error out of `firstWhere` and a listener with no `onError`.
- **Synchronous authenticator throws.** Authentication now goes through `Future.sync` inside the guarded block, so a throw before the future is returned gets the same refusal and the same cleanup (timer cancelled, subscription cancelled) as any other failure.
- **Build provenance.** Right: `--ref` only picks which copy of the workflow runs, and the Android build's checkout left `ref` unset, so a `release_tag` dispatch built `main`'s tip. Rather than soften the "built from" line, the build now checks out the tag it was dispatched for — every other trigger keeps checkout's default. That makes the recorded digests attributable to the tagged commit.

Full suite green (4252), analyze and format clean.


---

TheZupZup:
Closes #574 
Closes #575 
Closes #576

TheZupZup:
Third round, both findings fixed in `6e0b771`.

- **Every receiver command is gated, not just the handoff.** Play, pause, seek, volume, mute and status forwarded after the proved connection ended, which means a command reaching a receiver nobody authenticated — the delegate may have reconnected underneath, and this session's proof says nothing about that connection. They all go through the trust check now. Teardown stays ungated on purpose: closing an untrusted session is exactly what a caller should still be able to do. Test covers all seven commands refusing after a drop, and `close()` still working.
- **Old-tag rebuilds keep working.** Good catch: building at the release tag means an older tag's tree, which predates `verify_release_containment.py`, so the run would have finished the whole build and then failed on a missing file. The step now says plainly that nothing was checked, and why, instead of failing a rebuild on a script that tag never had. Every current release builds a tree that has it, so this only affects the documented rebuild path.

Full suite green (4253), analyze/format/ruff clean.


---

TheZupZup:
Fourth round, both fixed in `55fbfcb`.

- **The gate owns the wording now.** A typed `CastReceiverTrustException` from an authenticator kept its message all the way to `CastState.message`, so an implementation that built its message from something the receiver said would have put it on screen — with only a doc comment standing in the way. The refusal keeps its `CastTrustFailureKind` and nothing else; the gate writes the message per kind. Test: an authenticator refusing with a fingerprint and a certificate subject in its message, neither of which survives.
- **Linux packaging has the same old-tag guard as Android.** A tag predating the verifier would have built, packaged, and then failed on a missing script. Same handling, same warning.

Full suite green (4254), analyze/format/ruff clean.


---

TheZupZup:
Also fixed in `54d8f9b`: the publish workflow wrote "✅ published and verified" as soon as the assets existed, before the containment check on the downloaded bytes. A failed check then appended a failure section under a banner already claiming verification — exactly backwards for a security release's record. The presence step now only checks presence and hands on the Release URL; the banner and the digest table are written together by the containment step, after it passes.


---

TheZupZup:
Right, and my compatibility path was fail-open. Fixed properly in `70608f3`.

Rebuilding `v0.2.6` — a release that *does* carry the containment — would have replaced all five public Android assets (and the Linux tarball) with `--clobber` on the strength of a warning, just because the tag's tree has no copy of the verifier. That's exactly the case the check exists for.

The script and the expected message now come from different places, which is what the rebuild path needs: a second, script-only checkout of the workflow's own revision supplies the verifier, and a new `--containment-source` flag points it at the *built* tree's `cast_containment.dart`. So an older tag is verified against its own message rather than skipped.

The only remaining skip is a tag from before the containment existed at all, where there is no message to look for. That's decided by inspecting the built tree, not by whether the script file happens to be there.

Verifier unit tests cover the new flag (a matching older message passes, a mismatched one fails); ruff clean.


---

TheZupZup:
Fixed in `c33561d`. Same hazard as the close, one line further down: cancelling a stream subscription can wait on the stream's own teardown, so a delegate stalling there would have held the refusal inside the `finally` — `connect()` would never close the session or return the failure, and the UI would sit in "connecting" even though the authentication timeout had fired.

Both waits now go through one small bounded helper, and the trusted session uses it for its own lifetime subscription too. Test: a `readyStream` whose cancellation never completes still lets the refusal out.

Full suite green (4255), analyze and format clean.


---

TheZupZup:
Fixed in `e743770`. Same treatment for the successful session's own teardown, which is where it matters most: a receiver stalling while it closes would have held `DefaultCastService._teardownSession()` open, so a disconnect or session-loss recovery would never reach the idle state and the next connection would be stuck behind it. Test: closing returns even when the delegate's close never completes.

Full suite green (4256), analyze and format clean.


---

TheZupZup:
Two more, fixed in `08f858a` (and `main` merged in on top).

- **Synchronous cleanup failures.** `settleWithin` took a future, so the cleanup was started before the helper's `try`. An implementation that throws before returning one would have replaced the refusal the caller was already carrying. It takes a callback now and runs it through `Future.sync`. Test: a `close()` that throws before returning a future still lets the refusal out.
- **The Jellyfin lifetime claim was too comfortable, and that one matters.** I checked: `JellyfinSettingsController.clear()` forgets the token locally and never calls a server logout, so "reaches the whole account until the session ends" reads as "signing out takes it back" when it doesn't — a receiver that kept the URL keeps working until the user revokes that device in Jellyfin. Subsonic is the same shape (the credential is derived from the password). Both profiles say so now, and a test holds them to it: a model whose whole point is honesty can't carry the reassuring version.

Revoking server-side on sign-out is written up in `docs/cast-media-access.md` as its own change — it affects every user, not just those who cast, so it shouldn't ride along here.

Full suite green (4272 after the `main` merge), analyze and format clean.


---

Issue #408, opened 2026-08-16
[Linux] Read embedded album artwork from local files


TheZupZup:
Part of #376.

## Goal
Read embedded album artwork from local audio files on Linux and feed it through Linthra's existing artwork pipeline.

## Likely code areas
- `lib/core/sources/local/local_audio_metadata.dart`
- local artwork extraction/cache
- `lib/shared/widgets/artwork_image.dart`
- local metadata/artwork tests

## Requirements
- Extract common embedded cover-art tags without loading unnecessarily huge images into memory.
- Store extracted bytes in Linthra-managed private cache and persist only a safe local file reference.
- Reuse the same artwork rendering seam used by remote/local covers.
- Missing/corrupt artwork must fall back to the normal placeholder.
- Cache cleanup must never delete the user's source audio file.

## Acceptance criteria
- [ ] Embedded covers render in track/album/artist/Now Playing surfaces on Linux.
- [ ] Restart can reuse extracted artwork without reparsing every file when cache is valid.
- [ ] Missing/corrupt cache can be regenerated safely.
- [ ] Large artwork is bounded/resized appropriately.
- [ ] Fixture tests cover artwork present/absent/corrupt cases.

Borhan2004:
Working on this next: extract embedded cover art on Linux behind the existing LocalMetadataReader seam, cache bytes in Linthra-managed private storage (never the user's source file), and feed a safe file:// URI through Track.artworkUri.

TheZupZup:
## Findings preserved from PR #513

PR #513 explored and implemented a Linux embedded-artwork cache on top of the metadata-reader work from #512/#407:

- A private `linthra_local_artwork` cache can store extracted artwork instead of exposing or persisting the user's music-file path.
- SHA-1-derived cache keys provide stable cache paths.
- Atomic writes were used for cache safety.
- An 8 MiB artwork size cap was used to bound extraction/storage.
- Embedded-picture extraction can prefer front-cover artwork when multiple pictures are present.
- `LocalAudioMetadata.artworkUri` can point to a safe cached `file://` URI.
- Cache hits can skip image extraction (`getImage`) so rescans do not repeatedly parse large APIC/FLAC picture blocks.
- Missing, corrupt or oversized artwork falls back to the normal placeholder.
- Android SAF behavior remains unchanged.
- Unit tests for the artwork cache, filesystem metadata reader and platform bindings passed on the PR after formatting was fixed.

Still needs real Linux validation before closing this issue:
[...]

Issue #407, opened 2026-08-16
[Linux] Read local audio metadata natively


TheZupZup:
Part of #376.

## Goal
Replace the Linux local-music metadata fallback with a real desktop metadata reader.

## Likely code areas
- `lib/core/sources/local/local_audio_metadata.dart`
- local library scanner/source mapping
- local metadata abstractions
- metadata fixtures/tests

## Requirements
- Read title, artist, album artist, album, track number, disc number, year, genre and duration from common local audio formats.
- Keep implementation behind the existing metadata-reader seam so Android keeps its current path.
- Missing/corrupt tags must fall back safely to filename/existing defaults instead of dropping the track.
- Normalize optional fields consistently with current catalog models.
- Avoid loading entire large audio files into memory only to read tags.

## Acceptance criteria
- [ ] Tagged local files populate the expected Track fields on Linux.
- [ ] Untagged/corrupt files still appear safely.
- [ ] Multi-disc/album-artist fields are mapped correctly.
- [ ] Android metadata behavior is unchanged.
- [ ] Fixture-based tests cover representative formats and missing tags.

Borhan2004:
Working on this next: Linux-only metadata reader behind the existing LocalMetadataReader seam, with fixture-based coverage for tagged/untagged files and no Android behavior change.

TheZupZup:
## Findings preserved from reverted PR #512

The reverted implementation validated a workable native Linux metadata path behind the existing `LocalMetadataReader` abstraction:

- `audio_metadata_reader` (MIT) can be used on Linux through a `FilesystemLocalMetadataReader` implementation.
- The reader can map common ID3 / Vorbis / MP4 / APE / RIFF metadata into `LocalAudioMetadata`.
- Title, artist, album artist, album, track number and duration were handled in the reverted implementation.
- Missing or corrupt tags can safely fall back without throwing or dropping the track.
- Android can remain on its existing path; filesystem metadata reading does not need to alter Android SAF behavior.
- Embedded artwork should remain a separate concern (see #408) rather than coupling image extraction to the basic metadata path.
- Flatpak generated dependency sources need to be refreshed when adding the metadata-reader dependency.
- Relevant unit tests passed on the PR after a formatting fix.

Important gap discovered relative to this issue's full acceptance criteria: the reverted implementation summary did not claim complete mapping for every requested field (notably disc number, year and genre), so those should still be verified/implemented before closing this issue.

[...]

PR #590, opened 2026-09-09
feat(linux): read embedded local cover art (#408)

TheZupZup:
Refs #408. Follows on from #407 (#568), which deliberately left embedded artwork for this issue.

## The problem

Local Linux tracks got real titles/artists/albums from #407, but never a cover: `FilesystemLocalMetadataReader` always read with `getImage: false`, so `Track.artworkUri` stayed null for every local file and the library, album/artist tiles and Now Playing all fell back to the placeholder.

## What this adds

**`LocalArtworkCache`** (`lib/core/services/local_artwork_cache.dart`): extracts a local file's embedded cover (ID3 `APIC`, FLAC `PICTURE`, MP4 `covr`) into Linthra's private, OS-reclaimable cache and hands back a `file:` URI — the same seam `artworkImageProvider` already loads local covers through, so no changes were needed downstream.

- **Cache-first, so scanning stays cheap.** `FilesystemLocalMetadataReader` checks the cache *before* deciding whether to ask the parser for the embedded picture. A cache hit reads with `getImage: false`, exactly like a tags-only read; only a genuine miss pays for decoding the (possibly large) embedded image. This is what keeps #407's whole point — scanning a huge library doesn't get slower just because this landed.
- **Bounded, not just cached.** A cover wider or taller than 1024px is downsampled via `dart:ui` (decode + re-encode), preserving aspect ratio, before it ever reaches disk. A cover already within bounds is stored unchanged — no needless re-encode.
- **Never drops a track.** Corrupt or undecodable embedded bytes yield no artwork (not an exception), and text tags are extracted independently of the picture, so a good title next to a broken cover still shows the title.
- **Privacy-consistent with the rest of the cache layer.** The cache key is a SHA-256 of the *source file's own path*, never the path itself — mirrors the Android SAF walk's own embedded-artwork cache (SHA-1 of the content URI) and the pattern `ArtworkDiskCache`/`MediaArtworkCache` already use for their own keys.

## Tests

- `test/core/services/local_artwork_cache_test.dart`: store/cachedFile round-trip, distinct paths get distinct entries, the source path never appears in the cache directory, empty/undecodable bytes are rejected, a 0-byte cache entry is treated as a miss, resize bound (over/under/exact), and re-storing overwrites the same entry.
- `test/core/sources/local/filesystem_local_metadata_reader_test.dart`: MP3/FLAC extraction, no-picture-file has no artwork, picture-only file (no text tags) still reports it, corrupt embedded bytes don't affect text tags, a restart-equivalent fresh reader instance reuses the cache without rewriting the file, a deleted cache entry regenerates, and the resize bound end-to-end.
- Extended `audio_tag_fixtures.dart` with real ID3 `APIC` and FLAC `PICTURE` block builders (byte-level, matching the exact layouts `audio_metadata_reader`'s parsers walk), plus real decodable PNGs generated in-test via `dart:ui` for the resize assertions.

**Note:** I don't have a Flutter SDK available in the sandbox this was written in, so I could not run `flutter analyze`/`flutter test` myself — please let CI confirm, or run locally before merge.

### DIFF
--- a/lib/core/services/local_artwork_cache.dart
+++ b/lib/core/services/local_artwork_cache.dart
@@ -1,0 +1,138 @@
+import 'dart:convert';
+import 'dart:io';
+import 'dart:typed_data';
+import 'dart:ui' as ui;
+
+import 'package:crypto/crypto.dart';
+import 'package:path/path.dart' as p;
+import 'package:path_provider/path_provider.dart';
+
+/// Persists a local file's embedded cover art (ID3 `APIC`, a FLAC `PICTURE`
+/// block, an MP4 `covr` atom, …) into Linthra's private, OS-reclaimable cache
+/// and hands back a `file:` URI that [Track.artworkUri] can point at directly
+/// — [artworkImageProvider] already loads a `file:` URI straight off disk, so
+/// nothing downstream needs to know this cover came from a local file's own
+/// tags rather than a server.
+///
+/// Mirrors the Android SAF walk's own embedded-artwork cache
+/// (`SafDocumentScanner.cacheEmbeddedArtwork`): extract once, key by a hash of
+/// the *source file's path* (never the path itself, see [_key]) so a rescan
+/// reuses what an earlier one already extracted, write through a temp sibling
+/// then rename so a crash mid-write can never leave a half-written cover that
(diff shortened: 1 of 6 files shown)


TheZupZup:
Closes #408

PR #603, opened 2026-09-09
feat(downloads): download a whole album or playlist (#84)

### LINKED ISSUE #84: Add album / playlist "download all" actions
## Overview

Today downloads are per-track. Letting users download a whole album or playlist would be convenient — but it must stay true to Linthra's "no surprise downloads" promise.

## What needs to be done

- [ ] Add a "Download all" action for albums and playlists
- [ ] Show a confirmation before starting a bulk download
- [ ] Respect the existing cache size limit
- [ ] Respect the Wi-Fi-only / mobile-data setting
- [ ] Honour "Keep offline" / pinned behaviour so bulk downloads aren't silently evicted
- [ ] Guard against accidental **full-library** downloads

## Goal

Users can deliberately download an album or playlist for offline use, with no surprise or runaway downloads.

## Notes

- Downloads must remain **user-initiated** and Wi-Fi-only by default — this is a core privacy/UX promise.
- See [docs/offline-cache.md](./docs/offline-cache.md).

---

_Suggested labels: `enhancement`, `privacy`, `ui-polish`_

TheZupZup:
Closes #84.

Downloads have been per-track since the start. This adds an explicit **Download all** for one album or one playlist, without loosening any of the promises the offline cache already makes.

## What a user sees

- An album's app bar gets a download action; a playlist's menu gets **Download all**. Both appear only when something in that collection actually streams from a server, matching how the per-track menu hides offline actions for on-device files.
- It always asks first, naming how many songs the collection holds, and starts nothing until you confirm. The wording is about making the collection available offline rather than promising N downloads, because songs you already have are skipped.
- While a batch runs, the Downloads screen shows which album or playlist it is and how far along it is (`Downloading "Discovery": 12 of 30`) with a **Stop** button. The usual per-track rows sit below it, so cancel and retry for one song work exactly as before.
- When it ends you get one line saying what happened: everything offline, some queued (and what to do about it), some failed and retryable, some too large for the cache, older downloads evicted to make room, or you stopped it.

## How it works

This is not a second downloader. Every track goes through the existing `DownloadRepository.requestDownload`, so the Wi-Fi/mobile-data policy, the cache limit and its eviction, pinned downloads, per-track progress, dedup and per-track cancel all stay in the one place that already enforces them.

| File | Responsibility |
| --- | --- |
| `lib/core/services/bulk_downloader.dart` | Turns a collection into per-track requests. Skips what is already offline, collapses duplicates by provider-aware cache key, keeps a small number of requests outstanding so a batch stays stoppable, and reads the outcome back from the repository's own per-track status. |
| `lib/core/models/bulk_download_summary.dart` | Counts and flags only (never a URL, token or path), plus the one friendly completion line. |
| `lib/features/downloads/bulk_download_controller.dart` | Holds the single running batch and its stop flag. A second "Download all" is refused, not queued. |
| `lib/features/downloads/collection_download_actions.dart` | Filters the collection to what actually streams, deduplicates it, confirms, and starts the batch. |
| `lib/features/downloads/downloads_screen.dart` | The running-batch banner and its Stop button. |

Album and playlist screens only gained the entry point and a `canDownloadAll` check.

Two repository changes come with it, both in `CacheDownloadRepository`:

- Recording an on-device track and promoting a preloaded one now run inside the existing commit chain. `_save()` hands the store a snapshot taken at call time and finishes asynchronously, so two unserialized saves could land out of order and lose a cached-track record. Batching is what makes those paths concurrent routinely.
- The promotion re-reads its cache entry *inside* that commit. Waiting for the chain is exactly when a commit queued ahead can evict the preloaded copy and delete its file, and promoting the copy read before the wait would persist a `downloaded` record pointing at nothing. When it is gone the request falls through to a real download.

## Safety

- **No runaway downloads.** The action exists only where one specific album or playlist is open. There is no "download everything" entry point anywhere, the confirmation names the count, and a collection of 50+ gets an extra warning line.
- **Existing offline files are protected.** Tracks that are already downloaded are filtered out before any request, so re-running it on a mostly-downloaded album costs nothing and cannot disturb what is on disk.
- **On-device songs are left out.** A mixed playlist only sends its streaming tracks to the batch: local files are already on disk, and their rows deliberately offer no offline action.
- **Stopping never deletes.** A stop only ends the queue. Songs already downloaded stay downloaded, and the few requests in flight finish. Dropping one of those is still the per-track cancel.
- **The cache limit still wins, and says what it did.** A song the cache cannot fit is skipped and named; the batch carries on, because one oversized track should not cost you the rest of the album. Several refusals in a row means the cache really is out of room, so the batch stops rather than fetching more bytes it would throw away. If the limit evicted older downloads to make room, the summary says how many.
- **No secrets, and no promises it cannot keep.** Every string is built from counts and the collection name. A batch queued because you are offline says to start it again from Downloads, because nothing retries it for you.

## Tests

`test/core/services/bulk_downloader_test.dart` (18) covers one request per track, duplicate collapsing, two providers' same-id copies staying distinct, already-offline tracks never being re-requested, a partial failure not stopping the batch, an oversized track not stopping it either, repeated refusals stopping it, an evicted collection track not counted as still offline, a track downloading elsewhere not counted as failed, network-policy queueing (including a mid-batch change and the offline wording), cancellation, the outstanding-request bound, progress ticks, and eviction reporting.

`test/features/downloads/bulk_download_test.dart` (15) covers the album and playlist entry points and their gating, the confirmation and cancel path, the cache and Wi-Fi-only messages, a batch that cannot run at all leaving the app usable, a mixed collection requesting only its streaming tracks, songs already offline being counted honestly, duplicates staying consistent end to end, two concurrent starts running one batch, the Downloads banner and Stop, and a second batch being refused.

`test/data/repositories/cache_download_repository_test.dart` gains one: a preloaded copy evicted while its promotion waits on the commit chain is downloaded for real instead of restored as a record whose file is gone. It fails without that fix.

Ran locally with the pinned Flutter 3.44.7: `dart format --set-exit-if-changed .` clean, `flutter analyze` clean, `flutter test` 4833 passing.

## Limitations

- A queued batch does not resume by itself when the connection returns; the summary says to start it again from Downloads. Auto-resume would be a separate change to the repository. Related and pre-existing: the **single-track** queued message still promises an automatic start that nothing delivers. That is worth its own issue rather than widening this PR.
- A batch does not survive the app being killed. Songs already downloaded stay downloaded.
- `removeDownload`, `setPinned` and `notePlayed` still write cache metadata outside the commit chain. Same out-of-order-write shape as the two paths fixed here, but they were already reachable concurrently before this change, so they are left for a separate fix.
- The desktop right-click collection menu (`collection_menu.dart`) is shared with artists, so it deliberately did not get this entry. Adding it per-surface is a small follow-up.

## Docs

`docs/offline-cache.md` gains a "Downloading a whole album or playlist" section, `docs/codebase-tour.md` describes the new seam, `docs/manual-test-checklist.md` gains QA steps, and the stale "no download all yet" limitation is removed from `docs/jellyfin.md` and `docs/play-store-listing.md`.

https://claude.ai/code/session_01BsdW4hJnfxXeYfYexGWCuV


TheZupZup:
Thanks for the review, these were good catches. I checked all six against the code and every one of them was real. Fixed in d093e05, one summary here rather than six thread replies.

**1. Offline count after eviction.** Right, `offlineNow` was `alreadyOffline + downloaded`, so a collection track evicted to fit another song in the same batch still counted. The whole final accounting now reads the repository's per-track status snapshot instead of adding up request results, so `offlineNow` is what is actually offline when the batch ends. This also fixes #6.

**2. Cache refusal ending the batch.** Agreed, `CacheStorageException` also means "this one track is bigger than the free-able cache", which says nothing about a smaller track after it. One refusal no longer stops anything: refusals are counted and named in the summary ("1 was too large for the cache limit"), and the batch carries on. Only `maxConsecutiveCacheRefusals` (3) in a row ends it, which is the point where continuing would fetch bytes only to throw them away at commit. That bandwidth cost is why it does not just run to the end.

**3. Dedup vs the number shown.** Fixed by deduplicating earlier rather than by keeping two counts. `BulkDownloader.uniqueTracks` is now public and `CollectionDownloadActions` confirms exactly the set it will hand over, so `[A, A, B]` confirms 2, shows 2 in the progress denominator, and reports on 2.

**4. Unserialized metadata writes.** Confirmed: `_save()` hands the store a snapshot taken at call time and completes asynchronously, so two unserialized saves can land out of order and the older snapshot wins. Both request paths that were outside the commit chain (recording an on-device track, promoting a preloaded one) now run inside `_commit`, alongside the remote path. Batching is what makes those concurrent routinely, so it belongs in this PR. `removeDownload`, `setPinned` and `notePlayed` have the same shape and are left alone here: they were already reachable concurrently before this change and fixing them is not this PR's job.

**5. Controller reservation.** Correct, the guard only read published state and the first publish is an await away. `BulkDownloadController` now claims the slot synchronously before the first await and clears it in a `finally`.

**6. Track already downloading elsewhere.** Same fix as #1: a duplicate request returns "started" at once while the original fetch runs, and the track's status is `downloading`, not `failed`, so it is now counted as neither done nor failed. The Downloads screen shows it live.

Regression tests for each: an evicted collection track, an oversized track not stopping the batch, repeated refusals stopping it, a track downloading elsewhere not counted as failed, the duplicate count staying consistent end to end, and two concurrent starts running one batch. `dart format`, `flutter analyze` and `flutter test` (4829) all clean locally.

---

TheZupZup:
Second round, four more findings, all four real again. Fixed in 8e7e353.

**Preloaded promotion race.** This one is the best catch of the two rounds, and it is my own doing: serializing the promotion in the previous commit is exactly what created the window. The entry was captured before joining the chain, so a commit queued ahead could evict that copy and delete its file while this one waited, and the promotion would then persist a `downloaded` record pointing at nothing. The entry is now re-read inside the commit; when it is gone the request falls through to a real download.

There is a regression test for it, gated on `maxCacheBytes()` (the first await inside a commit) so the eviction lands in that exact window. Without the fix it fails with the record present and its file missing, which is the bug.

**Counting songs already offline.** Right, `unique.length` counted the whole collection while `run` drops what is already cached. Rather than querying offline keys from the UI layer, the wording now matches what the batch actually promises: "Make all 10 songs from X available offline?" instead of claiming 10 downloads, and the start snackbar drops its count entirely. The Downloads banner carries the real numbers, so nothing disagrees.

**Local files in mixed collections.** Confirmed, and it contradicted the gate right above it: `canDownloadAll` hides the action for an all-local collection precisely because those rows offer no offline action, then the batch registered them anyway. The collection is filtered through `isRemote` before anything is confirmed or requested. The test for it caught a second bug while I was there: the filtered list was being built but the unfiltered one was still handed to the controller.

**False automatic-retry promise.** Correct, and I had already listed the missing auto-resume as a limitation while still printing a message that promised it. The batch no longer reuses the single-track wording for the offline case; it says to start it again from Downloads. The Wi-Fi case keeps the repository's wording, which points at the setting that unblocks it and promises nothing.

Worth flagging separately: the **single-track** path still shows "This download will start automatically when you're back online", and it is equally untrue there, since nothing listens for a reconnect. That predates this PR and fixing it means either changing that message app-wide or implementing the resume, so it wants its own issue rather than being folded in here.

`dart format`, `flutter analyze` and `flutter test` (4833) all clean locally.

---

PR #610, opened 2026-09-10
test(flatpak): prove a chosen music folder works, and nothing else does (#447)

### LINKED ISSUE #447: [Flatpak] Add Flatpak local-library sandbox test
Part of #376.

## Goal
Validate the full local-library flow inside Flatpak: choose a folder, scan metadata/artwork and play a local track without broad host access.

## Likely code areas
- local folder portal/source setup
- local scanner/metadata reader
- Flatpak manifest permissions
- Linux playback controller
- sandbox test scripts/fixtures

## Requirements
- Use a disposable test music folder granted through the same portal/user-selection path as production.
- Verify scan, metadata, artwork and playback from that folder.
- Confirm an unrelated host path cannot be read without an explicit grant.
- Test revoked/missing access as a recoverable source error.
- Keep fixture media small and redistributable/test-safe.

## Acceptance criteria
- [ ] User-selected folder is scanned successfully in Flatpak.
- [ ] Local metadata/artwork load and a track plays.
- [ ] Unrelated host files remain inaccessible.
- [ ] Restart/revoked-access behavior is documented and safe.
- [ ] Test can be repeated locally and, where practical, in CI.

TheZupZup:
Closes #447.

#609 has merged, so this now targets `main` directly and the diff is only the local-library work plus a small tidy-up of the audio smoke's prose.

The Flatpak ships **no filesystem permission at all**. Local music works only because the user hands over one folder. That is a good policy and a fragile one: it holds only if the narrow path really works end to end, and nothing we ran checked that from inside the sandbox.

## The chosen folder works

| Assertion | Why it is not obvious |
| --- | --- |
| The folder scans | Without the grant it is not even visible from inside |
| Two tracks imported, one non-audio file **skipped** | A track count that happens to match proves less than a skip that is counted |
| Title, artist, album, track number come from the file's own tags | The filename fallback would produce plausible-looking values from nothing |
| The WAV reports a real duration | Only the container can supply that |
| The MP3's embedded cover is extracted, cached and decodes | Artwork is a separate I/O path with its own cache |
| The cover is cached **outside** the music folder | A scan must never write into the library it was handed |
| A track plays, position advancing | A status flag flipping is not playback |

Everything below the grant is the production wiring: the same `LocalLibraryScanner`, `LocalMusicSource`, `PlatformAudioFileScanner`, `FilesystemLocalMetadataReader` and `LinuxPlaybackController` that `LibraryController` assembles.

## Nothing else came with it

A sentinel file and a sibling directory sit beside the music folder in `$HOME`, and neither may be visible or readable from inside. The harness then re-checks both **on the host** afterwards, because "invisible" would prove nothing if the smoke had simply deleted them.

That pair is the sharp end of the policy: granting one folder inside `$HOME` must not hand over the rest of `$HOME`. It is now a test rather than a claim.

## Losing access is recoverable, not destructive

The last mode runs the same package against the same folder path with no grant at all, which is what a revoked portal document, an unplugged drive and a deleted folder all look like from inside. Linthra has to classify it as `folderUnavailable`, carry a user-facing message, **keep** the tracks already indexed for that folder, and refuse to persist the scan, so an unreadable folder can never be written down as "this library is empty now". With no catalog to fall back on it must report `retentionUnavailable` and still refuse.

## What stands in for the portal, and what does not

The grant is `flatpak run --filesystem=<folder>`: one folder, for one run, in no persistent override and in nothing the package declares. That is the *scope* a document-portal selection produces.

It is **not** the chooser dialog, and it does not exercise the document portal's `/run/user/<uid>/doc/…` rewriting. No headless runner can click that button. The manual portal pass in `flatpak-filesystem-audit.md` therefore stays, minus everything after the click, which is the part that breaks silently. `flatpak override --filesystem=…` is never used; the harness refuses to run when one already exists, since an override would decide the isolation result before the test started.

No `--filesystem=home`, no `--filesystem=host`, and nothing added to the manifest. A guardrail test asserts that on the submission manifest and on the harness's own commands.

## Fixture

Generated into a disposable `$HOME` directory and deleted on exit: a tagged WAV that really decodes (RIFF INFO has no picture field), an MP3 carrying an APIC cover, and a `notes.txt` that must be skipped. Tag bytes come from `test/core/sources/local/audio_tag_fixtures.dart`, the byte-level builders the local-metadata unit tests already use, imported rather than copied, so there is one place to be wrong about a container format. No committed media, no network, no credentials.

## Shared build

Rather than a third 90-minute `flatpak-builder` run, the generator from #609 now installs a *named set* of smoke binaries and one job runs both. Build artifacts are renamed `audio-smoke` to `sandbox-smoke` accordingly; the manifest equality proof is unchanged and still covers every added command.

## What was actually validated

**In CI, inside the installed Flatpak** ([run 34500460286](https://github.com/TheZupZup/Linthra/actions/runs/34500460286)):

```
PASS: scanned the selected folder, 3 files visited, 2 tracks imported, 1 skipped.
PASS: local metadata loaded from the file's own tags.
PASS: embedded artwork extracted and cached (64x64), outside the music folder.
PASS: played a track from the selected folder (position 202ms).
PASS: an unrelated host file stayed unreadable.
PASS: a sibling host directory stayed invisible.
PASS: the selected library survived a restart.
PASS: a revoked folder is a recoverable error
PASS: previously indexed tracks were kept, not wiped.
PASS: with nothing to retain, the scan still refuses to persist.
```

Both host probes were still present and readable on the host afterwards, so the invisibility was the sandbox rather than the test having deleted them.

Locally in this session against a real Flutter 3.44.7 SDK: `flutter analyze` clean, `dart format --set-exit-if-changed` clean, `flutter test test/tooling/` 455 passing, `python3 test/tooling/make_flatpak_smoke_manifest_test.py` 19 passing, `ruff check` and `ruff format --check` clean, `bash -n` on the harness, `scripts/check_secrets.sh` clean, `scripts/check_linux_runner.py` clean.

## Still manual

The chooser dialog itself. A portal grant is minted by a person clicking a button in xdg-desktop-portal, and no headless runner can click it. What is automated here is everything downstream of that click.

https://claude.ai/code/session_01QLDN6yzLpSdXoFHYJNqyyp


TheZupZup:
Green on `4064e5a`. [Run 34500460286](https://github.com/TheZupZup/Linthra/actions/runs/34500460286), first completed run of this smoke against the packaged app.

The whole sequence, from inside the installed Flatpak, with exactly one folder granted:

```
PASS: scanned the selected folder, 3 files visited, 2 tracks imported, 1 skipped.
PASS: local metadata loaded from the file's own tags.
PASS: embedded artwork extracted and cached (64x64), outside the music folder.
PASS: played a track from the selected folder (position 202ms).
PASS: an unrelated host file stayed unreadable.
PASS: a sibling host directory stayed invisible.
PASS: the selected library survived a restart.
PASS: a revoked folder is a recoverable error
PASS: previously indexed tracks were kept, not wiped.
PASS: with nothing to retain, the scan still refuses to persist.
```

A few things worth pulling out.

**The isolation result is the sandbox, not the test.** Both host probes sit in `$HOME` beside the music folder, and the harness re-reads them on the host after the sandbox runs. They were still there and still readable, so "invisible" means invisible rather than deleted.

**The revoked case produces a real message.** The error the scan surfaces is the one a user would see: "Linthra couldn't find the selected folder. It may have been moved or removed, the drive may not be connected, or access to it was revoked. Try selecting the folder again." That covers a revoked portal document, an unplugged drive and a deleted folder identically, which is correct, because from inside the sandbox they are identical.

**Losing access is not destructive.** The previously indexed tracks stay, and the scan refuses to persist an empty library over them. That was the failure mode worth testing: a sandbox permission lapsing should not look like "your library is now empty".

**The grant stayed narrow throughout.** One `--filesystem=<folder>` for the length of one run, no override, and the submission manifest still carries no filesystem permission at all. The harness refuses to run if an override exists, so a pass cannot belong to somebody's leftover `flatpak override`.

Still a documented manual step: the chooser dialog itself. A portal grant is minted by a person clicking a button in xdg-desktop-portal, and no headless runner can click it. What is automated here is everything downstream of the grant, which is the production path.

The audio smoke (#609) also ran green in this job, on `/app/lib/libmpv.so.2.5.0`, since both smokes share the one derived manifest and the one build.

---

Issue #456, opened 2026-08-16
[Flathub] Submit Linthra to Flathub


TheZupZup:
Part of #376.

## Goal
Submit Linthra to Flathub and track reviewer feedback until the app is accepted and publicly installable.

## Prerequisites
- Working reproducible Flatpak (#432–#448)
- Clean lint/submission metadata (#449–#455)
- Real Linux playback/local/server validation
- Final screenshots/release notes

## Likely code areas
- Flathub submission repository/PR
- upstream Flatpak manifest/metadata if reviewers find issues
- packaging docs

## Requirements
- Open the submission only after local/CI Flatpak builds and smoke tests are green.
- Respond to reviewer requests with small auditable upstream-first fixes.
- Do not weaken sandbox/security simply to accelerate acceptance.
- Keep the upstream issue linked to the submission so contributors can follow progress.
- Re-test the final accepted build, not only the pre-submission local build.

## Acceptance criteria
- [ ] Flathub submission PR is opened with validated files.
- [ ] Reviewer feedback is resolved or documented.
- [ ] Final package builds successfully on Flathub infrastructure.
- [ ] Linthra is publicly installable from Flathub.
- [ ] A clean install can launch and play validated content after acceptance.

TheZupZup:
Pre-submission blocker discovered while re-checking current Flathub policy (2026-09-03): Flathub's Generative AI policy now applies to the application itself as well as the submission PR, including code, documentation, manifests, metadata, patches and build scripts. It also says submission PR text/review replies must not be LLM-generated. Exceptions may be granted for mature, well-maintained projects.

Because Linthra has AI-assisted upstream work in its history, do not open the Flathub submission PR on the assumption that a human-written PR description alone is enough. Before #456 is executed, the maintainer should clarify/obtain an exception from Flathub or otherwise determine a compliant path.

This does not change the technical prerequisites: finish #449-#455, validate a stable immutable release, and keep the eventual Flathub PR itself human-authored/manual.

Issue #450, opened 2026-08-16
[Flathub] Complete AppStream metadata for submission


TheZupZup:
Part of #376.

## Goal
Bring Linthra's AppStream metadata from initial-valid to final Flathub submission quality.

## Likely code areas
- AppStream metainfo from #435
- screenshot/listing assets from #437
- release metadata/docs

## Requirements
- Verify summary/description accurately reflect the Linux app that is actually shipping.
- Add final project/developer/license links, content rating if required by the current submission rules, screenshots and release entries.
- Keep app ID, launchable desktop ID and version/release information synchronized.
- Avoid marketing claims for features not validated on Linux.
- Validate with current AppStream/Flathub tooling before submission.

## Acceptance criteria
- [ ] Metainfo contains all required final submission fields.
- [ ] Screenshots/releases are real and current.
- [ ] Version/app identity matches the Flatpak package.
- [ ] AppStream validation is clean.
- [ ] Metadata is ready to be consumed by a public software-center listing.

TheZupZup:
Current-state audit (2026-09-03): the AppStream file is structurally present and already has the app ID, licenses, developer, launchable, icon, categories, content rating and a 0.2.4 release entry. The remaining problem is accuracy/completeness, not initial creation.

Concrete drift to resolve before closing this issue:

- The current description says Linthra "plays local files" only.
- The XML comment still says Jellyfin/Navidrome/Subsonic/Plex are omitted until Flatpak network permission #440.
- #440 is now completed and the Flatpak manifest grants normal network access for configured self-hosted endpoints.
- Real Linux screenshots are still intentionally absent pending #437.

Update the public metadata only after the corresponding Linux workflows have been validated in the Flatpak. Keep claims limited to tested Linux behavior, add real screenshots, keep the AppStream release version/date synchronized with the stable release selected for Flathub, then run the Flathub-specific AppStream/linter validation (warnings count as failures there).

Issue #440, opened 2026-08-16
[Flatpak] Allow self-hosted server network access


TheZupZup:
Part of #376.

## Goal
Allow the Flatpak sandbox to reach user-configured Jellyfin, Navidrome/Subsonic and Plex servers without unrelated permissions.

## Likely code areas
- Flatpak manifest network permission
- provider HTTP/session clients
- sandbox integration tests/docs

## Requirements
- Grant only the network capability required for normal HTTP(S)/LAN access.
- Verify private/LAN addresses, hostnames, HTTPS and remote/tunnel endpoints work as normal network destinations.
- Do not add host filesystem or broad DBus permissions to solve networking.
- Provider discovery features, if any, should be evaluated separately from direct configured server access.
- Network denial/failure must surface through existing provider error states.

## Acceptance criteria
- [ ] Configured Jellyfin/Navidrome/Plex endpoints are reachable in Flatpak.
- [ ] Direct LAN and HTTPS server use are validated.
- [ ] No unrelated sandbox permission is added.
- [ ] Offline/network-denied state fails cleanly.
- [ ] Integration/manual test steps are documented.

Issue #437, opened 2026-08-16
[Flatpak] Add Flathub-quality screenshots


TheZupZup:
Part of #376.

## Goal
Create polished real screenshots for the Flatpak/Flathub listing once the desktop UI is representative.

## Likely code areas
- AppStream metainfo screenshot references
- `docs/` / listing assets
- desktop UI only if screenshot capture reveals obvious polish issues

## Requirements
- Capture the real desktop build, not stretched Android screenshots or mockups presented as shipped UI.
- Show useful views such as library/sidebar, album browsing, Now Playing/lyrics and self-hosted source experience.
- Use sample/test metadata that exposes no private server addresses, usernames, tokens or personal paths.
- Keep consistent window size/theme and readable content.
- Store source/README notes for repeatable future refreshes if practical.

## Acceptance criteria
- [ ] At least a small coherent set of desktop screenshots is ready.
- [ ] Images reflect current shipped behavior.
- [ ] No private information is visible.
- [ ] AppStream references point to stable assets when publishing.
- [ ] Screenshots remain legible at software-center preview sizes.

PR #623, opened 2026-09-10
chore(flathub): make the listing describe the Linux app that ships (#450)

TheZupZup:
Works on #450.

> **Stacked on #614.** Both touch `io.github.thezupzup.linthra.metainfo.xml`, and #614 carries the summary-length fix. Basing this on `main` would conflict on that file. GitHub will retarget it to `main` when #614 merges.

Most of #450's acceptance criteria were already met and I want to be plain about that rather than claim credit for them: `appstreamcli validate --pedantic` was already clean, the releases were already real and current, and app id, launchable and version were already synchronized (`scripts/check_release_metadata_sync.py` holds them there). What was actually wrong was smaller and more specific.

## The stale claim

The metainfo said streaming was left out "until the Flatpak has network permission (#440)". That permission is in the manifest, and #440 closed on 2026-09-03. Anyone reading the file would conclude networking was still blocked and that adding streaming to the listing was just waiting on packaging.

It is not. The real reason is that **nobody has signed in to a server from the installed Flatpak and played a track.** That is one of #456's own prerequisites and it is still open. So streaming stays out of the listing, with the honest reason written down.

## A guardrail, because this one is easy to get wrong

The network permission is exactly what makes a streaming claim a plausible accident: the code supports Jellyfin, Navidrome/Subsonic and Plex, there are unit tests, and the sandbox can reach the network. None of that is the same as having run it.

So the description may not name any of those providers, nor Chromecast, Android Auto or the share sheet. Both tests were verified to fail with the claim injected, and pass with it removed:

```
+14 -1: claims match what is validated on Linux no streaming provider is advertised
```

This is the rule from the issue ("do not claim a feature merely because unit tests exist") turned into something that fails a PR rather than something a reviewer has to remember.

## Screenshots stay missing, and #614 stays red

`metainfo-missing-screenshots` is a real linter finding. It gets no exception and I have not touched the exceptions file.

[`docs/flathub-screenshots.md`](https://github.com/TheZupZup/Linthra/blob/chore/450-appstream-metadata/docs/flathub-screenshots.md) is what closes it: which four views to capture and why each is defensible, what must not be on screen (no hostnames, usernames, tokens or personal paths, per #437), the window size and theme to hold constant, where the files go, and the exact `<screenshots>` block to paste.

Two judgement calls in there worth flagging:

**The captures should not come from CI.** I checked whether I could generate them: a headless runner draws through software EGL with no GPU and no compositor, and the result would look like a broken app rather than a real one. A listing screenshot is the one asset where "technically the shipped UI" is not good enough. They want a real GNOME or KDE session.

**The URLs should pin to a tag, not a branch.** `raw.githubusercontent.com/.../main/...` would let a later commit silently change the images behind a published listing. A tag cannot move, so the listing keeps showing what was reviewed.

## Also

Added `<keywords>`, which is what a software centre actually searches on. A listing without them is much harder to find, and getting them right costs nothing.

## What was validated

Locally, against a real Flutter 3.44.7 SDK and AppStream 1.0.2:

- `appstreamcli validate --pedantic` on the metainfo, clean
- `flutter test test/tooling/` 452 passing, including the 16 in `flathub_metadata_guardrails_test.dart`
- the two new guardrails verified to fail with a false claim injected, then restored
- `flutter analyze` clean, `python3 scripts/check_release_metadata_sync.py` clean, `python3 scripts/check_linux_runner.py` clean, `scripts/check_secrets.sh` clean

**Not run here:** the real `flatpak-builder-lint`. It needs a Flatpak and a full build, so the linter's own verdict comes from CI. It will still report the missing screenshots, which is the correct outcome for this PR.

https://claude.ai/code/session_01QLDN6yzLpSdXoFHYJNqyyp

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLDN6yzLpSdXoFHYJNqyyp)_


TheZupZup:
CI is red here and will stay red until the screenshots exist. Recording why rather than leaving it looking unattended.

**Failing check:** Build and launch Flatpak, on the `flathub_builder_lint.py` step ([job](https://github.com/TheZupZup/Linthra/actions/runs/34506364526/job/102969430466)).

```
--- repo flatpak/repo-ci ---
  appstream-screenshots-not-mirrored-in-ostree (error)
  metainfo-missing-screenshots (error)
```

Both are inherited from #614, which this is stacked on, and both are the correct outcome. `metainfo-missing-screenshots` is a real submission blocker, not noise, and it gets no entry in `flathub-lint-exceptions.json`. Suppressing it would defeat the point of #449 having added the linter at all.

**I am not fixing it in this PR, and not widening it to try.** The captures have to come from a real desktop session. I checked whether CI could produce them and it can't do so honestly: a headless runner draws through software EGL with no GPU or compositor, so the images would misrepresent the app in the one place people judge it. TheZupZup is taking them; [`docs/flathub-screenshots.md`](https://github.com/TheZupZup/Linthra/blob/chore/450-appstream-metadata/docs/flathub-screenshots.md), added in this PR, is the spec.

No re-run: this is deterministic, not a flake.

Two things worth pulling out of the same log, because they are the useful signal:

**The build succeeded.** `Flatpak build and repository export completed successfully`. The two findings come from the `repo` mode inspecting the exported OSTree, not from anything failing to build.

**The `appstream` mode is clean:**

```
--- appstream .../io.github.thezupzup.linthra.xml.gz ---
  no findings
  message: Validation was successful.
```

That is this PR's own change being checked. The keywords and the rewritten rationale survive `appstreamcli compose` into the app-info XML that Flathub actually consumes, which is a stronger statement than the local `appstreamcli validate --pedantic` I ran. So the metadata work here is verified; only the screenshots are outstanding.

Once the PNGs land and the `<screenshots>` block goes in, both findings clear together and #614 goes green with it.

---

Issue #536, opened 2026-08-31
media server always connected


oniGino:
# Linthra bug report

## Summary
media server always connected

## What happened
even when traveling app falsely reports media server as connected. there is no way so exclude this unavailable music from my library without fully clearing it from connections

## Steps to reproduce
1. 
2. 
3.

## Expected behavior
when remote dont show me music from jellyfin

## Diagnostics
```text
Linthra diagnostics
App version: 0.2.4
Android: 2026081301
Jellyfin: connected
Jellyfin host: 192.168.22.1:8096
Library tracks: 9063
Local folder: selected
Local folder access: persisted
Local supported types: aac, flac, m4a, mp3, ogg, opus, wav
Cache: 417.5 MB of 4 GB
Playback output: local
Playback state: error
Current track: id#15087424
Playback at last background: loading
Last lifecycle: resumed
Notification permission: granted
Last error: none
Last interruption: resolution
Cast available: yes
Cast connected: no
Android Auto supported: yes
Offline cache: enabled
```

## Recent app events
```text
precache: start:3
lifecycle: inactive
bg-playback: loading
lifecycle: hidden
bg-playback: loading
lifecycle: paused
bg-playback: loading
error: resolution
lifecycle: hidden
bg-playback: error
lifecycle: inactive
bg-playback: error
lifecycle: resumed
precache: start:3
error: resolution
precache: start:3
lifecycle: inactive
bg-playback: loading
lifecycle: hidden
bg-playback: loading
lifecycle: paused
bg-playback: loading
lifecycle: detached
lifecycle: resumed
lifecycle: inactive
bg-playback: loading
lifecycle: resumed
error: resolution
```

Issue #431, opened 2026-08-16
[Desktop] Add offline/server availability indicators


TheZupZup:
Part of #376.

## Goal
Make desktop library surfaces show whether content is local, cached/offline-capable, online, unavailable or reconnecting when that information is useful.

## Likely code areas
- `lib/features/library/` track/album rows
- download/offline cache state
- provider availability/session state
- unified logical-track model/presentation helpers
- widget tests

## Requirements
- Derive indicators from existing source/cache state; do not add provider-specific network checks per row.
- Keep status compact and understandable; avoid visual noise on every item when everything is normal.
- Distinguish explicitly downloaded/cached audio from merely remote/streamable content.
- Never expose provider URLs/tokens.
- Status must update when cache/provider availability changes.

## Acceptance criteria
- [ ] Offline/local/remote/unavailable states are represented consistently.
- [ ] Indicators reflect actual cached/downloaded state.
- [ ] Provider reconnect updates status without full app restart.
- [ ] Unified-library rows remain source-safe and secret-free.
- [ ] Representative status combinations have widget/provider tests.

Borhan2004:
@TheZupZup 

I’d like to take this issue.

Apologies for the delayed response — I’m back to working on Linthra and would like to pick this up now.

I’ll first review the existing library, cache/download, and provider availability state to make sure the implementation fits the current architecture rather than introducing unnecessary provider-specific checks.

I’ll cover the required offline/local/remote/unavailable/reconnecting states, ensure the status reacts to cache and provider availability changes, and add the necessary widget/provider tests for the representative combinations.

Borhan2004:
@TheZupZup  Can you please upgrade the flutter version of your repository ? 

Because the repo pins Flutter 3.44.7 but I have 3.47.2 installed.

TheZupZup:
Hey @Borhan2004, it's good now 👍
I've upgraded the repository to the latest stable Flutter version, so you should be good to continue working on #431.
Thanks again for pointing out the outdated Flutter version!

PR #630, opened 2026-09-11
feat(library): show whether a track row still plays with its server a…

Borhan2004:
…way (#431)

A library row already carried a download glyph, but that only describes an in-flight *task*. It said nothing about the question a server-backed row raises and a local one never does: will this still play when the server is away? The availability state needed to answer it already exists (#536) — the per-source probe controller publishes it, and the download repository maintains the offline cache set — but no library surface read it, so a row whose server had just gone away looked exactly like a row whose server was fine.

This adds the missing presentation layer, deliberately quietly:

* `TrackAvailabilityStatus` + `resolveTrackAvailability` (pure, total): every combination of "what this device has" and "what the server is doing" resolves to one answer, including the ones that are awkward to stage on screen.
* `TrackStatusGlyph` owns the row's single status glyph across both axes — the download task and the settled availability — so a row can never show an in-flight ring next to a contradictory availability icon.
* The row stays silent by default: on-device music, and server-backed music whose server is answering and which has no saved copy, render nothing at all.

Guarantees:

* Derived, never probed. The inputs are the existing `sourceAvailabilityProvider` and `offlineAvailableTrackKeysProvider`; no row performs a network check.
* Live. State is read through Riverpod, so a server that comes back updates every visible row with no restart, no reconnect and no rescan.
* Secret-free. The only text announced is a fixed server name from `PlaybackSourceLabel` ("Jellyfin", "Navidrome", "Plex"); a URL, host, username, token or path is unreachable from the glyph.
* Explicit downloads, auto-prefetched copies and streamable-only rows are distinguished, reusing the distinction the records already keep (`CachedTrack.preloaded`) rather than inventing one.
* Scoped per source, so a probe of one server never rebuilds another's rows, and an idle row adds no subscription.

Untouched: no probe, resolver, playback path or download-policy change — a source that starts publishing availability gets its indicators for free, and every unlisted source is still treated as reachable. Album and artist rows are deliberately not covered yet: an album mixes tracks, so "available" would first have to mean something specific.

Tests: the resolver's full state table, the glyph's states and a11y labels, the no-noise guarantee, the one-glyph-never-two rule, and a reconnect that updates an on-screen row without a restart.

### DIFF
--- a/docs/library.md
+++ b/docs/library.md
@@ -153,6 +153,61 @@ track or surfaced in the Library — a Jellyfin track's stored reference is an
 opaque `jellyfin:<id>`, and stream URLs (which carry the access token) are minted
 only at play time.
 
+## Track status in the library
+
+Every track row carries one small, non-interactive glyph beside its overflow
+menu. It never grows into a control, and it is **silent by default**: on-device
+music has no server to be away from, and a server-backed row whose server is
+answering and which has no saved copy of its own has nothing worth saying. An
+indicator lit on every ordinary row is one nobody reads, so the interesting
+states earn their place by being the exception.
+
+The glyph answers two questions, in this order:
+
+1. **Is something being transferred?** Queued, downloading (a determinate ring
+   when the server reported a size), or failed. Transient and actionable, so it
+   wins whenever it has something to say.
+2. **Will this still play when the server is away?** Once nothing is in flight,
+   the row reports its settled availability.
+
+| What the row is | Glyph | Meaning |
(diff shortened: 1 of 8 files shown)


TheZupZup (approved):

TheZupZup:
Really nice to see you back, Borhan!

I reviewed the PR and it looks solid. Thanks for taking the time to fit it into the existing availability/cache architecture instead of adding another parallel path.

Glad to have you contributing to Linthra again 🙌

Issue #446, opened 2026-08-16
[Flatpak] Add Flatpak audio playback smoke test


TheZupZup:
Part of #376.

## Goal
Prove the packaged Flatpak can initialize the real Linux audio backend and complete a short playback lifecycle inside the sandbox.

## Likely code areas
- `tool/linux_audio_backend_smoke.dart`
- Flatpak test/launch scripts
- manifest/libmpv runtime packaging
- `.github/workflows/` when automated

## Requirements
- Reuse the existing native audio lifecycle smoke where practical instead of inventing a second backend test.
- Run the smoke from the installed Flatpak environment so missing libmpv/plugins/runtime files are caught.
- Use a tiny local fixture; no real server credentials or public network.
- Validate initialize → load → play/pause/seek/stop/dispose as supported by the existing smoke.
- Capture useful sanitized failure output.

## Acceptance criteria
- [ ] Smoke passes from an installed Flatpak with bundled runtime.
- [ ] Removing/breaking libmpv makes the test fail for the right reason.
- [ ] No host libmpv dependency can mask packaging errors.
- [ ] Test remains credential/network independent.
- [ ] Local and CI instructions are documented.

PR #634, opened 2026-09-11
fix(flatpak): the audio and local-library smokes delete only what they made (#629)

### LINKED ISSUE #629: The audio and local-library Flatpak smokes can delete a contributor's app data
`scripts/flatpak_audio_smoke.sh` and `scripts/flatpak_local_library_smoke.sh` both end their cleanup with:

```bash
flatpak --user uninstall -y --delete-data "$APP_ID"
```

Both refuse to start if the app is already installed, and the comment in the audio smoke leans on that ("Safe because this script refused to start against an installation it did not make"). That reasoning has two holes.

### 1. The data tree outlives an ordinary uninstall

Somebody who removed an older Linthra build without `--delete-data` still has `~/.var/app/io.github.thezupzup.linthra/` with their library database, settings, offline audio and credentials in it. Neither script is installed at that point, so both will happily run, and then wipe the tree on the way out.

`docs/flatpak-development.md` already calls that flag **Destructive** and spells out exactly what it removes, so this is the repo's own documented hazard being triggered by the repo's own tooling.

### 2. `--delete-data` also clears the permission store

It removes the app's entries from Flatpak's permission store, which lives outside `~/.var/app` entirely. That is where the document-portal grants live: the music folders a user picked through the file chooser. So even a check on the data directory does not cover what the flag actually does, and a user with grants but no data tree loses them silently.

Narrower than the first problem (a user re-picks their folders; nothing of theirs is deleted), but it is the same mistake: a guard about one thing authorising an action that reaches past it.

It only bites locally. CI runners start clean, so nothing is at risk there, which is also why it has gone unnoticed.

### Fix

Check whether the data tree exists *before* installing, and remove only a tree this run created, directly rather than through the flag:

```bash
APP_DATA_DIR="$HOME/.var/app/$APP_ID"
APP_DATA_EXISTED=0
[[ -e "$APP_DATA_DIR" ]] && APP_DATA_EXISTED=1
```

then in cleanup, plain `uninstall` always, followed by `rm -rf -- "$APP_DATA_DIR"` only when `APP_DATA_EXISTED` is 0. That way the guard and the action are about the same thing, and the permission store is never touched.

`scripts/flatpak_filesystem_smoke.sh` was fixed this way in #615 (Codex found both halves there first). The audio smoke needs one extra thought: its `--delete-data` also doubles as the cleanup for the negative control's shadow directory, so that needs handling on its own when the tree is pre-existing.

Worth doing at the same time: `docs/flatpak-audio-smoke.md` and `docs/flatpak-ci.md` both tell contributors to run these locally, and neither warns about it today.

TheZupZup:
Closes #629

## The failure mode

Both harnesses ended their cleanup with:

```bash
flatpak --user uninstall -y --delete-data "$APP_ID"
```

Run either one on a machine that has `~/.var/app/io.github.thezupzup.linthra/`, and the tree goes: library database, settings, offline audio, credentials. CI never noticed because runners start clean. Locally it only takes one contributor reproducing a packaging failure.

## Why "is Linthra installed?" was not enough

Both scripts refuse to start against an installed Linthra, and the audio smoke's comment leaned on that ("Safe because this script refused to start against an installation it did not make"). Two holes:

1. **The data tree outlives an ordinary uninstall.** Somebody who removed an older build without `--delete-data` still has the whole tree, and `flatpak info` reports nothing installed, so the guard passes and the cleanup wipes it.
2. **`--delete-data` reaches past the tree.** It also drops the app's entries from Flatpak's permission store, which lives outside `~/.var/app` and holds the document-portal grants for the music folders a user picked. A check on the data directory could not have authorised that even if it had existed. Those grants are what the local-library smoke exists to prove work.

So the guard was about the installation, and the action reached into data and grants the run never created.

## The cleanup ownership model

Same shape `scripts/flatpak_filesystem_smoke.sh` has had since #615:

- before installing anything, record whether `$HOME/.var/app/$APP_ID` exists (`APP_DATA_EXISTED`). It has to be asked first, because after the install the answer is always yes;
- on exit, always `flatpak --user uninstall -y "$APP_ID"`, never `--delete-data`;
- `rm -rf` the tree directly, and only when this run created it;
- the permission store is never touched by either script.

The audio smoke needed one more thing, as the issue called out. Its negative control writes a shadow libmpv into the sandbox's cache, which the old flag cleaned up as a side effect of deleting the whole tree. The directory name is now a single constant (`SHADOW_DIR_NAME`), passed into the sandbox and read by the host cleanup as `$APP_DATA_DIR/cache/$SHADOW_DIR_NAME`, so a run that leaves a pre-existing tree alone still removes that directory by name. The control now requires `XDG_CACHE_HOME` instead of falling back to `$HOME/.cache`, so the path the sandbox writes and the path the host cleans are provably the same one; a sandbox without it exits 3, which is the existing "could not set this control up" failure.

Nothing about what the controls prove changed: all three sonames are still shadowed with a library that loads, the shadow control is still `bounded`, the identity control is still `named`, and an exit-0 control still fails the job.

## What stops it coming back

New: `test/tooling/flatpak_smoke_cleanup_test.py`, wired into `ci.yml` (cheap, offline, so it runs on every PR rather than only in the 90-minute Flatpak job where this never bites).

It runs **both real harnesses end to end** against stub `flatpak`, `xvfb-run` and `dbus-run-session` binaries in a throwaway `HOME`, so these are behavioural, not textual:

- a tree the run created is gone afterwards (CI behaviour is unchanged);
- a tree that was there first survives, with its library file intact, and so does a stand-in permission-store entry under `~/.local/share/flatpak/db/`;
- the negative control's shadow directory is created inside that tree and then removed on its own;
- the fixture music folder, sibling folder and sentinel are cleaned up;
- every `uninstall` the run issued carried no `--delete-data`.

The stub honours `--delete-data` the way flatpak does (tree plus permission store), so putting the flag back fails these tests rather than passing them quietly. I checked that by reintroducing it and by no-oping the shadow removal: four tests fail on the first, one on the second.

Alongside them, text rules:

- nothing under `scripts/` or `.github/workflows/` passes `--delete-data` (comments about the flag are excluded, commands are not);
- for each smoke that removes app data, every `rm -rf -- "$APP_DATA_DIR"` sits inside an `if`/`elif` carrying `! APP_DATA_EXISTED`;
- the guard is read before the first `flatpak --user install`;
- the docs that tell contributors to run these locally still describe what a run may delete.

`flatpak_audio_smoke_test.dart` and `flatpak_local_library_smoke_test.dart` both asserted `--delete-data` was present. They now assert the opposite, plus the new cleanup shape and the shadow-directory handling.

## Docs

- `flatpak-audio-smoke.md` and `flatpak-local-library-smoke.md` both promised a run removes app data unconditionally. They now say what it actually removes, and that a pre-existing tree and your portal grants are left alone.
- `flatpak-ci.md` said nothing about it, and it is the page that tells people to run the smokes locally. It gets a short "What a smoke may delete on your machine" section.
- `flatpak-development.md`'s uninstall table now names the permission store on the `--delete-data` row, which was the half of the flag it did not mention.

## Checks

`flutter test` (5309 tests, all passing), `flutter analyze`, `dart format --set-exit-if-changed`, `ruff check`/`ruff format --check`, `python3 test/tooling/flatpak_smoke_cleanup_test.py`, `scripts/check_flatpak_permissions.py`, `scripts/check_linux_runner.py`, `scripts/check_secrets.sh`. `shellcheck` on both scripts reports only the pre-existing SC2016 infos that `main` already has (the in-sandbox scripts are deliberately single-quoted).

CI on this head is green, including `Sandbox smokes on the packaged app`, which runs both changed harnesses against a real built package.

## One thing I left alone

`docs/linux-desktop.md` says "Uninstalling removes the application and its sandboxed data" two lines above explaining that `--delete-data` removes the data. That is the same misunderstanding in user-facing docs, but it is not smoke tooling, so it is out of this PR's scope. Happy to fix it in a one-line follow-up if you want.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,18 @@ jobs:
       - name: Test the Flatpak permission checker
         run: python3 test/tooling/check_flatpak_permissions_test.py
 
+      # A smoke borrows a contributor's machine: it may remove what it
+      # installed, and nothing else (#629). The audio and local-library smokes
+      # used to end with `uninstall --delete-data`, which deletes
+      # ~/.var/app/<app id>/ and the app's permission-store grants whether or
+      # not the run created them. These tests run the real harnesses against
+      # stub flatpak binaries in a throwaway HOME, so the rule is checked on
+      # every PR rather than only on a runner with flatpak-builder, where it
+      # never bites because the runner starts clean. Local twin:
+      # `python3 test/tooling/flatpak_smoke_cleanup_test.py`.
+      - name: Test what a Flatpak smoke may delete
+        run: python3 test/tooling/flatpak_smoke_cleanup_test.py
+
   release-bump-guard:
     name: Release bump touches only version files
     runs-on: ubuntu-latest
--- a/docs/flatpak-audio-smoke.md
+++ b/docs/flatpak-audio-smoke.md
@@ -159,8 +159,23 @@ bash ../scripts/flatpak_audio_smoke.sh repo-sandbox-smoke
 ```
 
 The harness refuses to run if Linthra is already installed, so it can never
-uninstall or test a build you care about. Everything it does install, it
-removes on exit, app data included.
+uninstall or test a build you care about.
+
+It also never runs `flatpak uninstall --delete-data`. That flag would take
+`~/.var/app/io.github.thezupzup.linthra/` with it (library database, settings,
+offline audio, credentials) and clear the app's entries in Flatpak's permission
+store, where the document-portal grants for the music folders you picked live.
+Neither of those is necessarily something the run created: an install you
+removed earlier *without* `--delete-data` leaves both behind, while
+`flatpak info` reports nothing installed at all.
+
+So what the harness cleans up is what it owns. On exit it removes the temporary
+remote, the install, and the app-data tree **only if that tree was not already
+there** when it started. If it was, the tree stays, and the one thing inside it
+this run made, the negative control's shadow directory under `cache/`, is
+removed by name instead. Your permission-store grants are never touched either
+way. A run does leave the packaged app's own settings and cache written into a
+tree you already had, the same as launching the app would.
 
 Every sandbox run is time-bounded (`LINTHRA_FLATPAK_SMOKE_TIMEOUT`, 300s by
 default), including the probe that checks the smoke binary is installed. The
--- a/docs/flatpak-ci.md
+++ b/docs/flatpak-ci.md
@@ -124,6 +124,30 @@ is already installed for the current user or system-wide, so a local smoke can
 never launch or remove a contributor's existing installation. Use a clean test
 user/environment for this final step when Linthra is already installed.
 
+### What a smoke may delete on your machine
+
+CI runners start clean, so none of this matters there. It matters locally, and
+the rule is the same for every smoke in `scripts/`: each removes what it
+created, and nothing else.
+
+None of them uses `flatpak uninstall --delete-data`. That flag deletes
+`~/.var/app/io.github.thezupzup.linthra/` (library database, settings, offline
+audio, credentials) and clears the app's Flatpak permission-store entries, the
+document-portal grants for the music folders you picked, and it reaches both
+whether or not the run put them there. An ordinary uninstall leaves a data tree
+behind, so "Linthra is not installed" says nothing about whether one exists.
+
+The three smokes that install the app themselves (`flatpak_audio_smoke.sh`,
+`flatpak_local_library_smoke.sh`, `flatpak_filesystem_smoke.sh`) therefore
+look for
+`~/.var/app/io.github.thezupzup.linthra/` *before* they install anything,
+uninstall plainly on the way out, and remove that directory only when it was
+not already there. `flatpak_launch_smoke.sh` never removes app data at all. So
+a data tree you already had, and your permission-store grants, are never
+removed by a smoke. For the commands that *do* delete that data, when you want
+it gone, see
+[flatpak-development.md § Clean and uninstall](./flatpak-development.md#clean-and-uninstall).
+
 ## The audio smoke job beside it
 
 The same workflow runs a second job, `Audio lifecycle smoke in the sandbox`,
--- a/docs/flatpak-development.md
+++ b/docs/flatpak-development.md
@@ -231,7 +231,7 @@ Ordered least to most destructive. Nothing here needs `sudo`.
 | `rm -rf flatpak/flatpak-builder-build flatpak/repo` | build tree + local repo | Safe; both are recreated by the next build. Delete the remote too, or it dangles |
 | `flatpak --user uninstall --unused` | runtimes nothing installed needs any more | Safe, but re-downloads them next time you build |
 | `rm -rf flatpak/.flatpak-builder` | downloaded sources + every cached module build | **Expensive.** The next build recompiles ffmpeg/libplacebo/libass/mpv and re-downloads the Flutter SDK |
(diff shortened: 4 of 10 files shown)


TheZupZup:
The `Security surface guard` check on this head is not going to finish, and the cause is not in this diff.

The guard runs `scripts/check_pr_security_surface.py` and scans the full post-change text of every file the PR adds lines to. One blocked-addition rule, "download-and-execute shell pattern", backtracks catastrophically on `test/tooling/flatpak_audio_smoke_test.dart`:

```
_SHELL_QUOTED          = (?:'(?:[^']*)'|"(?:\\.|[^"\\])*")
_SHELL_COMMAND_SEGMENT = (?:_SHELL_QUOTED|\\\n|[^\n|;&])*
```

`[^\n|;&]` also matches `'` and `"`, and `'(?:[^']*)'` can span newlines, so every quote after the first `curl`/`wget` token gives the engine two ways to consume the same text. Cost grows exponentially with the amount of text after that token. The file has carried `expect(harness, isNot(contains('curl ')));` since #446, at line 96 of 271.

Measured locally (Python 3.11, that one rule, against the file's full text):

| input | time |
| --- | --- |
| 60 lines after the token | 0.03 s |
| 120 lines | 2.1 s |
| 176 lines (main's copy, to EOF) | 25.8 s |
| this PR's copy (336 lines) | still running after 20 minutes |

So main's copy already costs 26 seconds for a single rule, and the 65 lines this PR adds to that file tip it over. Any PR that adds a line to that file hits the same wall.

One line fixes it:

```diff
-_SHELL_COMMAND_SEGMENT = rf"(?:{_SHELL_QUOTED}|\\\n|[^\n|;&])*"
+_SHELL_COMMAND_SEGMENT = rf"(?>(?:{_SHELL_QUOTED}|\\\n|[^\n|;&])*)"
```

An atomic group, available in `re` from Python 3.11 on, which is the floor here (the guard's own test workflow pins 3.11; the guard job uses the runner's `python3`). Verified locally:

* both copies of that file scan in 0.001 s;
* `test/tooling/check_pr_security_surface_test.py` (145 tests) and `check_pr_security_surface_edge_regression_test.py` (9 tests) pass unchanged;
* detection is the same on the obvious cases. Still matched: `curl … | sh`, `wget -qO- … | bash`, `curl -H "a|b" … | sudo bash`, a pipe across a line break, `sh <( curl … )`, `bash -c "$(curl …)"`, `eval "$(curl …)"`. Still not matched: `curl -o out.tar …`, `curl … || true`, and the `isNot(contains('curl '))` line above;
* the whole scan of this diff then finishes in under a second and reports sensitive, not blocked (`ci.yml` plus the two automation scripts), which passes the gate for an owner-authored PR.

I deliberately have not put that change in this PR. It is security tooling and unrelated to #629, and it would not unblock this PR anyway: the guard loads the scanner from the trusted base commit, not from the head, so the fix has to be on `main` before a re-run here can pass. Happy to open it as its own small PR if you want it.

Everything else on this head is green, including `Sandbox smokes on the packaged app`, which runs both changed harnesses against a real built package.

---

TheZupZup:
Correction to the comment above: the guard did finish. It passed at 00:12:44 UTC after 43m51s on the current head (`0703c39`, after the main merge), so this is a 44-minute tax on the gate, not a hard block, and every check on this head is green.

The measurements stand. The rule is exponential in the amount of text after the first `curl`/`wget` token in a scanned file, main's copy of `test/tooling/flatpak_audio_smoke_test.dart` alone costs about 26 seconds, and the next few dozen lines anyone adds to that file take the job from 44 minutes toward the 6-hour limit. The one-line atomic-group change still brings it to 0.001 s with the guard's own 145 + 9 tests passing, and it still belongs on `main` rather than in this PR.

---

Issue #391, opened 2026-08-16
[Desktop] Add configurable keyboard shortcuts


TheZupZup:
Part of #376.

## Goal
Add desktop keyboard shortcuts for common playback and navigation actions, with remapping where practical.

## Likely code areas
- app-level Actions/Shortcuts handling
- `lib/features/player/`
- navigation/search state
- settings persistence and tests

## Requirements
- Defaults for play/pause, next/previous, search, library, queue and Now Playing.
- Do not hijack text-editing shortcuts while an input has focus.
- Keep shortcut definitions centralized so UI/help can read the same source.
- Persist user overrides safely and handle conflicting bindings clearly.
- Media keys remain MPRIS territory (#398), not duplicate global shortcuts.

## Acceptance criteria
- [ ] Defaults work from normal desktop surfaces.
- [ ] Typing/search fields keep expected editing shortcuts.
- [ ] Remapped shortcuts persist across restart.
- [ ] Conflicts/invalid bindings are rejected or explained.
- [ ] Shortcut dispatch has focused tests.

PR #656, opened 2026-09-17
feat(desktop): add a keyboard shortcuts help window (#392)

### LINKED ISSUE #392: [Desktop] Add a keyboard shortcuts help window
Part of #376.

## Goal
Add a discoverable desktop help window listing Linthra's current keyboard shortcuts.

## Likely code areas
- desktop help/settings UI
- shared shortcut registry from #391
- widget tests

## Requirements
- Group shortcuts by playback, navigation and library actions.
- Read from the same shortcut definitions used by the app so the help cannot drift.
- Show user-remapped bindings when configuration exists.
- Window/dialog must itself be keyboard navigable and easy to close.

## Acceptance criteria
- [ ] Help view lists all supported desktop shortcuts.
- [ ] Remapped bindings appear correctly.
- [ ] No duplicated hard-coded shortcut table is introduced.
- [ ] Focus/escape behavior works without a mouse.
- [ ] Widget test verifies representative entries.

TheZupZup:
Closes #392

`Ctrl+/`, or **Show all shortcuts** on the Keyboard shortcuts settings card, opens a window listing every shortcut Linthra answers, grouped, with the combination each one is bound to right now.

### It has no list of its own

Rows come from the registry #391 added and chords from the live binding map the dispatcher installs. A remap shows up in the window without anything being kept in step by hand, live, while it is open, and so does a reset. A test holds every row to that map rather than to a written-out list, because a second hard-coded table is the exact failure this is meant to avoid.

### Grouping is part of the registry

`ShortcutGroup` is a field on the action, so "which heading does this go under" cannot drift from the action either. Headings follow the enum's declaration order and rows follow the registry's, so the window is deterministic and does not depend on map iteration. A group nothing is filed under is not drawn as an empty heading.

| Group | Actions | Defaults |
| --- | --- | --- |
| Playback | Play / pause, Next track, Previous track | `Ctrl+Space`, `Ctrl+→`, `Ctrl+←` |
| Navigation | Queue, Now Playing | `Ctrl+U`, `Ctrl+P` |
| Library | Search, Library | `Ctrl+K` (also `Ctrl+F`), `Ctrl+L` |
| Help | Shortcut help | `Ctrl+/` |

Search sits under Library rather than Navigation because what it searches is the library.

`Ctrl+F` is shown under search's binding as "or Ctrl + F" rather than on a row of its own, since it is a second way to press the same shortcut and not a second shortcut to learn.

### The chord goes through the registry like any other action

`Ctrl+/` is a `ShortcutAction`, which makes it listed, remappable, and impossible to bind something else over. `Shift+/` (the "?" people also reach for) is not offered: Shift plus a printable key is a capital letter, which `ShortcutBinding` refuses outright. Pressing it again while the window is open does nothing rather than stacking a second copy, which is worth guarding because the window lists its own chord.

### Keyboard and focus

Escape closes it through the `DismissIntent` every modal route already answers, rather than a second path to the same pop. The focus stop that opens with it sits *inside* the scroll view, which is what makes Page Up/Down and `Ctrl`+arrow scroll the list: Flutter's `ScrollAction` looks for a `Scrollable` above whatever holds the keyboard, so a stop outside it would leave those keys doing nothing. Tab moves on to Close. Closing hands the keyboard back to the control that opened it when there is one to name (the settings button passes its own node, because Flutter never focuses a button that was clicked), and otherwise to whatever held it when the chord fired.

### Nothing on the phone changes

The entry point is on the desktop-only settings card, and the one line pointing at that card is left out where the card does not exist. The window itself is not gated on the platform, the same rule the shortcuts themselves follow, so a tablet with a keyboard case gets it for free.

Both halves of a row wrap, so a long or translated action name grows the row taller instead of pushing the chord off the window.

### Tests

`test/features/help/keyboard_shortcuts_help_test.dart` covers the defaults appearing, a stored override appearing instead, a remap and a reset updating the window while it is open, deterministic grouping, every row matching the live map, Escape, Tab, scrolling, focus restoration from both entry points, a vanished opener, and narrow windows. `test/app/shortcuts/shortcut_action_test.dart` covers the grouping being total and deterministic and the new action's default. `test/app/shortcuts/linthra_shortcuts_test.dart` covers the chord dispatching, not stacking, and opening on a remap. `test/features/settings/desktop/keyboard_shortcuts_section_test.dart` covers the settings entry point.

`dart format`, `flutter analyze` and `flutter test` all clean. `docs/linux-desktop.md` updated.


TheZupZup:
`Validate simulated unlocked state` is red on `3100ebb`, and it is not this PR's failure.

Gradle could not resolve `org.jetbrains.kotlin:kotlin-gradle-plugin`, because `repo.maven.apache.org` answered HTTP 403 for the Kotlin plugin BOM and POMs:

```
> Could not get resource '.../kotlin-gradle-plugins-bom-2.2.20.pom'.
   > Could not GET 'https://repo.maven.apache.org/maven2/.../kotlin-gradle-plugins-bom-2.2.20.pom'.
     Received status code 403 from server: Forbidden
```

It died 32 seconds in, during dependency resolution, before anything was compiled.

Why it is not mine: this branch changes ten files, all Dart and Markdown, with no Gradle, Android, Kotlin or pubspec changes. The same check passed on the previous head `ae5624f` with identical app code, and the only thing that has changed since is the `main` merge. On this very head the two sibling Gradle jobs, `Validate simulated locked state` and `Build debug APK`, resolved the same Kotlin plugins and carried on building, so this was a transient 403 on one runner rather than a repository-wide outage.

There is no fix to port. Nothing upstream addresses a one-off Maven 403, and adding retries would mean editing `.github/workflows/github-sponsor-simulation-apk.yml`, which is shared CI config well outside the scope of #392.

I tried to re-run the failed job and the API returns 403 for `rerun-failed-jobs`, so I do not have the permission. A maintainer re-run should clear it. I am keeping the PR watched until it is green.

---

Issue #418, opened 2026-08-16
[Player] Add artist queue actions


TheZupZup:
Part of #376.

## Goal
Add desktop-friendly artist queue actions using Linthra's existing logical library and queue behavior.

## Likely code areas
- artist detail/list widgets under `lib/features/library/`
- playback/queue providers
- artist grouping/sorting logic
- tests

## Requirements
- Actions: Play artist, Play next, Add to queue.
- Define deterministic ordering (for example album/year/track or existing library ordering) instead of relying on map iteration.
- Preserve logical-track provider fallback behavior.
- Do not duplicate tracks that the unified library intentionally collapses.

## Acceptance criteria
- [ ] Artist actions enqueue the expected logical tracks exactly once.
- [ ] Ordering is documented and deterministic.
- [ ] Play next/appending preserve the existing queue correctly.
- [ ] Empty/unavailable artist states are safe.
- [ ] Queue/grouping tests cover representative artists.

Issue #85, opened 2026-05-26
Complete local metadata parsing with disc numbers


TheZupZup:
## Overview

Local metadata parsing is now largely implemented: Linthra reads title, artist, album artist, album, track number, duration, and embedded artwork for user-selected local music through Android's Storage Access Framework, with filename/folder fallbacks when tags are missing.

The remaining gap from the original issue is **disc number** support so multi-disc albums can be ordered and grouped correctly.

## What needs to be done

- [ ] Read disc number from local audio metadata when present
- [ ] Carry the disc number through `LocalAudioMetadata` and the local track mapping path
- [ ] Preserve the existing fallback behaviour when the tag is missing or malformed
- [ ] Add tests for multi-disc files and missing/partial disc metadata
- [ ] Keep file access SAF-scoped; do not add broad storage permission

## Goal

Local multi-disc albums preserve their disc ordering while keeping the metadata parsing and privacy model already shipped in Linthra.

## Already implemented

- [x] Title
- [x] Artist
- [x] Album artist
- [x] Album
- [x] Track number
- [x] Duration
- [x] Embedded artwork
- [x] Graceful filename/folder fallbacks
- [x] Scoped Storage Access Framework access

See `lib/core/sources/local/` and the Android SAF scanner for the current implementation.

Issue #449, opened 2026-08-16
[Flathub] Run flathub-builder-lint cleanly


TheZupZup:
Part of #376.

## Goal
Run Flathub's builder linter against Linthra's final manifest/metadata and reach a clean submission-quality result.

## Likely code areas
- Flatpak manifest/modules
- AppStream metadata
- desktop file/icons
- packaging helper scripts/docs

## Requirements
- Run the linter exactly against the files intended for submission.
- Fix actionable errors/warnings in the manifest, sources, permissions and metadata.
- Do not suppress findings merely to make the command green; document any justified exception.
- Re-run after permission/source/metadata changes that affect packaging.
- Keep this issue focused on lint/submission readiness, not unrelated UI work.

## Acceptance criteria
- [ ] Current submission files pass the linter with no unresolved actionable error.
- [ ] Any accepted exception has a written reason.
- [ ] Lint command is documented/repeatable locally.
- [ ] CI can run the same check where practical.
- [ ] Result is captured before opening the Flathub submission.

TheZupZup:
Current-state audit (2026-09-03): the upstream Flatpak CI already validates the desktop file/AppStream metadata, builds the generated Flatpak manifest from declared sources on a clean runner, exports an OSTree repo, installs the package, and runs the launch smoke test. What is still missing for this issue is the actual Flathub linter pass against the submission-shaped manifest and exported repo.

Before closing #449, run the current Flathub Builder commands against the files intended for submission:

- `flatpak run --command=flatpak-builder-lint org.flatpak.Builder manifest <manifest>`
- `flatpak run --command=flatpak-builder-lint org.flatpak.Builder repo repo`
- optionally the dedicated AppStream lint as well.

Do not treat the existing `appstreamcli validate` step as equivalent to the full Flathub linter. Capture every actionable finding, fix it upstream where appropriate, and only then consider wiring the same lint into CI. No linter exceptions should be added just to make the check green.

TheZupZup:
The manifest lint started failing on 2026-09-17, and it is currently red on every branch. Recording it here rather than on a PR, since it belongs to this issue and no PR can fix it in passing.

```
--- manifest flatpak/io.github.thezupzup.linthra.yml ---
  runtime-update-available-to-org.gnome.Platform-51 (warning)
  info: ['runtime-update-available-to-org.gnome.Platform-51: Please consider
         updating to the latest runtime version']
FAIL: 1 finding(s) with no written reason.
```

[...]

PR #657, opened 2026-09-17
feat(player): add album queue actions (#417)

### LINKED ISSUE #417: [Player] Add album queue actions
Part of #376.

## Goal
Add clear desktop album actions for starting or extending the playback queue.

## Likely code areas
- album detail/list widgets under `lib/features/library/`
- queue/playback controller/providers
- album grouping/order logic
- tests

## Requirements
- Actions: Play album, Play next, Add to queue.
- Preserve disc/track ordering and the logical-track/provider fallback model.
- Actions must use existing queue APIs rather than manually mutating player state.
- Handle empty/partially unavailable albums gracefully.

## Acceptance criteria
- [ ] Play album starts the correct ordered album queue.
- [ ] Play next inserts after the current track without destroying the rest of the queue.
- [ ] Add to queue appends deterministically.
- [ ] Multi-disc ordering is correct.
- [ ] Queue behavior has focused tests.

TheZupZup:
Album surfaces could only replace the queue. Play and Shuffle were the whole album page, so there was no way to line an album up behind what was already playing without throwing that away first.

The album page now offers all three actions. Play and Shuffle stay the header buttons they always were, and the two that extend the queue sit in a menu beside them: the same entries, and the same commands, the Albums tab's album menu already offers on right-click.

## What each action does

| Action | Behaviour |
| --- | --- |
| **Play** | Replaces the queue with the album and starts it. Normal play semantics, unchanged. |
| **Play next** | Inserts the whole album, in album order, directly after the current track. Everything already upcoming stays behind it. |
| **Add to queue** | Appends the whole album, in album order, to the end. |

From silence, Play next and Add to queue start the album at its first track instead of queueing it behind nothing, which is what the single-track versions already do.

All four run through the one shared `runCollectionAction`, over the very list the page is rendering. Nothing is re-derived for the queue and no widget touches queue internals. That list comes from the de-duplicated library, so a song held on two servers is queued once as the preferred copy (fallbacks intact), and a track whose server is away is simply not there to queue. An album with nothing playable does nothing at all.

## Two changes underneath

**Batched queue commands.** `PlaybackController` grows `playNextAll` / `addAllToQueue`, backed by `PlaybackQueue.enqueueAllNext` / `appendedAll`. A loop of single-track `playNext` calls queues a collection backwards (every insert lands at the same slot) and publishes a state change per track. One call inserts once, in order, and publishes once.

**Album order sorts by disc, then track.** Track numbers restart on every disc, so ordering by number alone interleaves a two-disc album (d1t1, d2t1, d1t2, ...). Nothing populates `Track.discNumber` yet, so the tier is a no-op today and single-disc albums order exactly as before; multi-disc albums fall into place as soon as #85 carries the tag through the mapping path. I kept that deliberately out of this PR so it stays a queue-actions change.

Artist actions are #418 and are not touched here.

## Tests

- Play album order, and that Play replaces an existing queue
- multi-disc order, on Play and on Play next
- Play next over an existing upcoming queue, and from silence
- Add to queue appends deterministically
- empty album (no actions offered, and the shared runner never reaches the controller)
- partly unreachable album queues the part that can play
- a song held on two providers is queued once, best copy first
- each album track queued exactly once
- one controller call and one published state per action
- the album page and its grid card produce the same queue

`flutter analyze`, `dart format --set-exit-if-changed .`, `flutter test` (6140 tests) and `./scripts/check_secrets.sh` all pass locally.

`docs/queue.md` picks up a short section on the album versions and a manual-checklist line.

Closes #417


TheZupZup:
**Build and launch Flatpak** is red here, and it is not this PR's.

The failure is the Flathub manifest lint, before anything is built:

```
--- manifest flatpak/io.github.thezupzup.linthra.yml ---
  runtime-update-available-to-org.gnome.Platform-51 (warning)
FAIL: 1 finding(s) with no written reason.
```

Same failure, byte for byte, on `main` at `c72965e` (the base of this branch) about 35 minutes before this run: [run 35183838746](https://github.com/TheZupZup/Linthra/actions/runs/35183838746). Flathub published GNOME Platform 51, so the linter started reporting that our `runtime-version: '50'` pin is behind. Nothing in this diff touches `flatpak/`.

No fix for it exists on any open branch yet, so there is nothing to port into this PR. Not re-running the job either: the linter compares against the live Flathub runtime list, so it will report the same thing until the pin moves.

The fix belongs in its own PR, under #449. Roughly:

- `runtime-version: '51'` in both `flatpak/flatpak-flutter.yml` and the generated `flatpak/io.github.thezupzup.linthra.yml`, then `./scripts/regenerate_flatpak_sources.sh`
- re-check the reasoning that is written against the 50 runtime and freedesktop-sdk 25.08: the libsecret/libgcrypt note, the gnutls.pc one, the libpipewire fallback and the GTK 3 icon theme path, plus whether `org.freedesktop.Sdk.Extension.llvm20` is still the right extension on the new base
- a full Flatpak build and launch to confirm it, since a runtime bump is exactly the change CI's smoke tests exist for

An entry in `flathub-lint-exceptions.json` would turn the check green, but the policy in that file rules it out: a finding that is simply not fixed yet belongs in its issue, not there. Agreed, so I have left it alone.

Happy to pick up the bump as a separate PR if you want it now.

---

